PR #32, opened 2026-02-09
(chore): added otel tracing to envoy proxy config

Koufan-De-King:


Dericko681 (approved):
pls add this patch to complement the mcp integrations since we want to isolate the mcps to unique ns  `kubectl -n default patch gateway ai-gateway --type='json' -p='[{"op":"replace","path":"/spec/listeners/0/allowedRoutes","value":{"namespaces":{"from":"All"}}}]'
`

PR #108, opened 2026-04-01
Change sync schedule to run at 2 AM

Franck-Sorel:
Updated the schedule time for two database sync configurations.

### DIFF
--- a/charts/apps/values.yaml
+++ b/charts/apps/values.yaml
@@ -1029,7 +1029,7 @@ applications:
                   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
               spec:
                 immediate: true
-                schedule: "0 2 * * *"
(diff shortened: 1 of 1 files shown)


charts/apps/values.yaml
```diff
@@ -1046,7 +1046,7 @@ applications:
                   argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true
               spec:
                 immediate: true
-                schedule: "0 2 * * *"
+                schedule: "0 0 2 * * *"
```
stephane-segning: Same

stephane-segning (changes requested):
I'm not sure this exactly was the issue. Plus the proposed schedule is not using a valid cron. I used https://crontab.guru to check.

Koufan-De-King (approved):

PR #256, opened 2026-04-24
Feat/wildcard tls

Franck-Sorel:


charts/apps/values.yaml
```diff
@@ -1446,7 +1446,7 @@ applications:
                 annotations:
                   cert-manager.io/cluster-issuer: cert-home-cert-http
                 host: "coder.ai.camer.digital"
-                wildcardHost: "serverless.coder.ai.camer.digital"
```
stephane-segning: Please use the domain we spoke about earlier

charts/cert/values.yaml
```diff
@@ -18,4 +18,23 @@ cert-manager:
   config:
     apiVersion: "controller.config.cert-manager.io/v1alpha1"
     kind: "ControllerConfiguration"
-    enableGatewayAPI: true
\ No newline at end of file
+    enableGatewayAPI: true
+
+issuer:
```
stephane-segning: The pattern is bad. Actually, I think it should be done another way to integrate this. Why call this `issuer` has no meaning. Call it `issuerCloudflare` instead

stephane-segning (changes requested):

stephane-segning (approved):

PR #271, opened 2026-05-05
269 document custom headers from librechat passed to upstream

### LINKED ISSUE #269: Document custom headers from librechat  passed to upstream
after testing through the pr #268  document it

benie-joy-possi:


Koufan-De-King (changes requested):
One document is a complete duplicate of the other, I don't know which is meant to be which.

Also, an example of one header being added would strengthen the message of this doc. Especially if the value of the header can come from keyclaok.

Koufan-De-King (approved):

PR #322, opened 2026-05-21
fix(gateway):optimize envoy retries logging and scaling

benie-joy-possi:


charts/core-gateway/templates/backendtrafficpolicy.yaml
```diff
@@ -11,7 +11,7 @@ spec:
   connection:
     bufferLimit: 100Mi
   retry:
-    numRetries: 5
+    numRetries: 3
```
Koufan-De-King: let's use 1 first and see

Koufan-De-King (approved):

PR #661, opened 2026-07-15
feat: add GitLab integration support with project-specific access tokens and webhook secrets

leghadjeu-christian:
## 1. Summary

This PR makes the dashboard fully GitLab-aware: repository links, run target deep-links,
trigger labels, and the admin approval view all now render the correct platform-specific
URLs and copy instead of hard-coding GitHub paths.

**Core fixes:**

- `Platform` enum was serialising as `"GitLab"` (PascalCase) — the frontend `=== "gitlab"` check never matched and every GitLab repo fell through to the GitHub default, producing 404 links. Fixed with `#[serde(rename_all = "lowercase")]`.
- `repo_platform` column was missing from both `list_tasks` and `get_task` SQL projections — `sqlx::FromRow` would fail at decode time even though the field is `Option<Platform>`. Both queries updated; regression test added.
- `repoUrl()` / `targetUrl()` in `tasks.ts` and `repos.ts` now dispatch on `repo_platform`: GitLab repos link to `/-/merge_requests/{iid}` / `/-/issues/{iid}`; GitHub repos keep `/pull/` and `/issues/`.
- `triggerLabel()` renders `MR #N` for GitLab and `PR #N` for GitHub.
- `github_repo_id` field (renamed to `platform_repo_id` by migration `0024_platform_abstraction`) is now reflected in all consumers. The `lci` CLI carries `#[serde(alias = "github_repo_id")]` for zero-downtime backward compatibility.

**Architecture:**

Instead of a single `GITLAB_URL` env var in Next.js, the control plane exposes a new
authenticated `GET /config` endpoint that returns `gitlab_base_url` + a per-project URL
map (`gitlab_project_base_urls`). This lets multi-host deployments (where different
GitLab projects live on different instances) resolve the correct base URL per project.
Server Components fetch this once at render time and pass a `GitlabLinkConfig` value
down — no client-side env vars, no `NEXT_PUBLIC_*` leakage.
---

It solves:
#653 

## 2. Intent

> GitLab runs were linking to `https://github.com/…` because `Platform` serialised as
> `"GitLab"` (PascalCase) and the frontend's `=== "gitlab"` check never matched,
> silently falling through to the GitHub default. Fixing the serde attribute alone was
> enough to unblock links; the rest of the PR hardens the URL-building for self-hosted
> and multi-host deployments.

---

### In Scope

**Rust — control-plane**
- `platform.rs`: add `#[serde(rename_all = "lowercase")]` to `Platform` enum
- `db/tasks.rs`: add `repo_platform: Option<Platform>` to `TaskRow`; include `r.platform AS repo_platform` in `list_tasks` and `get_task` SQL projections
- `db/tests.rs`: regression guard — `task_queries_include_repo_platform` verifies both queries decode the column correctly
- `http/config.rs` *(new)*: `GET /config` endpoint returning `{ gitlab_base_url, gitlab_project_base_urls }` derived from the GitLab registry; authenticated, no secrets exposed
- `http/mod.rs`, `main.rs`: wire `GET /config` into the router and application state

**TypeScript — web**
- `lib/domain/gitlab-links.ts` *(new)*: `GitlabLinkConfig` interface, `normalizeGitlabBaseUrl`, `gitlabLinkConfig`, `gitlabBaseUrlForProject` — per-project URL resolution with trailing-slash normalisation
- `lib/domain/tasks.ts`: `repo_platform` field on `Task`; `repoUrl`, `targetUrl`, `triggerLabel` become platform-aware; both accept `GitlabLinkConfig`
- `lib/domain/repos.ts`: `platform` + `platform_repo_id` fields on `Repository` (replaces the now-redundant `github_repo_id`); `repoUrl` uses per-project URL lookup
- `lib/server/api.ts`: `getDeploymentConfig()` fetch + `DeploymentConfig` type
- `lib/utils/config.ts`: `gitlabBaseUrl()` removed from env path; config now flows from the control plane via `GET /config`
- `components/repos/repo-list.tsx`: accepts `GitlabLinkConfig`; `RepoCard` shows "View on GitLab" / "View on GitHub" dynamically
- `app/dashboard/repositories/page.tsx`, `app/dashboard/runs/[id]/page.tsx`: fetch deployment config; construct `GitlabLinkConfig`; inject into child components
- `app/dashboard/admin/page.tsx`: shows `{platform} id {platform_repo_id}` instead of hardcoded `github id`

**Rust — lci CLI client**
- `clients/lci/src/api.rs`: `RepositoryRow.github_repo_id` → `platform_repo_id`; `#[serde(alias = "github_repo_id")]` preserves backward compat; fixture test extended to verify both field names deserialise correctly
- `clients/lci/src/tui/`, `render.rs`: struct literal field renames

### Out of Scope

- GitLab webhook ingestion changes (already handled in a prior PR)
- Frontend unit tests for `gitlab-links.ts` (follow-up ticket)
- `React.cache()` wrapping for `getDeploymentConfig` (follow-up optimisation)

---

## 4. Verification

- [x] Running automated tests
- [x] Running manual lint/build
- [ ] Running manual tests
- [ ] Checking logs
- [ ] Checking metrics
- [ ] Testing error cases
- [ ] Testing permissions/security behaviour

```bash
# TypeScript — zero errors
pnpm --filter web tsc --noEmit

# Rust — zero warnings, zero clippy lints
cargo fmt --check
cargo clippy          # clean across all crates
cargo check -p lci    # Finished dev profile, exit 0

# Control-plane DB tests (requires Postgres)
cargo test -p control-plane db::tests::task_queries_include_repo_platform
```

---

## 5. Risk Assessment

- [x] Low
- [ ] Medium
- [ ] High

**Potential risks:**

| Risk | Mitigation |
|---|---|
| Old task rows have `NULL` `repo_platform` (LEFT JOIN miss) | `Option<Platform>` + `default:` branch in `switch` falls through to GitHub — links hidden gracefully, no crash |
| GitLab disabled / `GET /config` fails | `gitlabLinkConfig(null, null)` → `normalizeGitlabBaseUrl(null)` → `"https://gitlab.com"` — SaaS fallback always works |
| `lci` CLI deployed before server redeployed | `#[serde(alias = "github_repo_id")]` accepts both field names |
| Multi-host GitLab (different projects on different instances) | `gitlabBaseUrlForProject(config, project_id)` resolves per-project; falls back to default |

---

## 6. AI Usage Declaration

AI was used for:

- [x] Understanding existing code
- [x] Generating code
- [x] Reviewing the diff

Human verification:

---

## 7. Reviewer Focus

- [x] Correctness — `Platform` serde fix, SQL projection, URL path format
- [x] Edge cases — `null` platform, self-hosted URL, multi-project multi-host
- [x] Backward compatibility — `serde(alias)` on `platform_repo_id`
- [x] Security — `GET /config` is authenticated, exposes only web URLs (no tokens)


charts/lightbridge-code-intelligence/values.yaml
```diff
@@ -267,6 +267,28 @@ config:
...
+  # actual secret values stay in Kubernetes env-injected secrets (never in the ConfigMap). The
+  # corresponding env vars must be injected into the control-plane pod from Kubernetes secrets (see
+  # the control-plane controller `env` block). Example:
+  #   gitlab:
+  #     enabled: true
+  #     defaultApiUrl: "https://gitlab.com/api/v4"
+  #     defaultBotHandle: "lightbridge-bot"
+  #     projects:
+  #       - projectId: 12345
+  #         accessToken: "{env:GITLAB_ACCESS_TOKEN_12345}"
```
stephane-segning, line 271: **The example's `{env:…}` marker names don't match the env vars this PR wires.** This example uses `{env:GITLAB_ACCESS_TOKEN_12345}` / `{env:GITLAB_WEBHOOK_SECRET_12345}` (keyed by project ID), but the control-plane container (values.yaml ~line 404) only injects `GITLAB_ACCESS_TOKEN_PROJECT_1` / `GITLAB_WEBHOOK_SECRET_PROJECT_1`. Since `{env:…}` resolves at runtime against the container env, a user copying this example gets an **unresolved marker → empty access token**, and GitLab API/webhook auth fails silently. Pick one scheme and make both halves agree — e.g. document the marker as `{env:GITLAB_ACCESS_TOKEN_PROJECT_1}` to match what's injected, or make the injected env-var names follow the `_<projectId>` convention shown here.

charts/lightbridge-code-intelligence/values.yaml
```diff
@@ -374,6 +396,23 @@ lci:
                 secretKeyRef:
                   name: '{{ .Release.Name }}-github'
                   key: webhook-secret
+            # GitLab per-project secrets (file-config-only, ADR-0072). The {env:…} markers in
+            # control-plane.json resolve these env vars at runtime via lci-config substitution.
+            # The `<release>-gitlab` Secret must be provisioned with the real values before
+            # GitLab webhooks/API calls work (see externalsecret.yaml for the pattern).
+            # Placeholder values here for chart defaults; overridden per-env in ai-helm-values.
+            GITLAB_ACCESS_TOKEN_PROJECT_1:
```
stephane-segning: **Single-project wiring under a "multi-project" feature.** These env vars are hardcoded to one project (`PROJECT_1` ← secret keys `access-token-1`/`webhook-secret-1`), but `config.gitlab.projects` renders an arbitrary-length list. A 2-project config produces `{env:…}` markers for project 2 that resolve to nothing — there's no `GITLAB_ACCESS_TOKEN_PROJECT_2` — so only the first project can ever authenticate. If multi-project is in scope, the env wiring needs to be generated per project (loop over `config.gitlab.projects`) rather than a fixed single entry; if it's deliberately one-project-for-now, say so in the comment and the values example so it isn't mistaken for working multi-project support.

charts/lightbridge-code-intelligence/values.yaml
```diff
@@ -374,6 +396,23 @@ lci:
...
+                secretKeyRef:
+                  name: '{{ .Release.Name }}-gitlab'
+                  key: access-token-1
+                  optional: true
+            GITLAB_WEBHOOK_SECRET_PROJECT_1:
+              valueFrom:
+                secretKeyRef:
+                  name: '{{ .Release.Name }}-gitlab'
+                  key: webhook-secret-1
+                  optional: true
```
stephane-segning: I'm not seeing where these ENVs are used. control-plane.json do not contains that, right?

stephane-segning (changes requested):

charts/lightbridge-code-intelligence/values.yaml
leghadjeu-christian: The `GITLAB_ACCESS_TOKEN_PROJECT_1`, the `PROJECT_1` is just a draft  example to use to open this pr while waiting for the real project ids and webhook secrets.

charts/lightbridge-code-intelligence/values.yaml
leghadjeu-christian: Same here, the template takes multiple projects `{{-   $projects := list }}`

charts/lightbridge-code-intelligence/values.yaml
leghadjeu-christian: They are used in the ai helm values [here](https://github.com/ADORSYS-GIS/ai-helm-values/pull/90/changes#diff-7ff0f5ff87814c74bdfb2728db5d3e98aa8ae76b69804aef73ad13e70ef6ad33R280-R283)

stephane-segning (approved):

PR #857, opened 2026-07-31
feat(model-serving): Deploy coding model to GPUs, replacing Qwen-VL

rustnew:
## 1. Summary

This PR changes:

- Disables `qwen3-vl-4b-thinking` (vision-language model) on the GPU fleet — frees one card for the replacement
- Adds `qwen3-coder-30b-a3b` — Qwen3-Coder-30B-A3B (MoE 128 experts, 8 active, AWQ INT4) on vLLM — the new coding model
- Enables FP8 KV cache as fleet default (`defaults.kvCacheDtype: fp8_e4m3fn`) for all vLLM models, with per-model override and 4 fail-fast validation guards in `_helpers.tpl`
- Removes the `qwen3-vl-4b-thinking-local` backend from `charts/ai-models/values.yaml` and sets its model entry to `enabled: false` (gateway off before server off, per `replace-a-model.md`)

It solves:

- Closes #836 — Deploy coding model to GPUs, replacing Qwen-VL

---

## 2. Intent

The intent of this PR is:

> Replace the non-coding-optimized Qwen3-VL-4B-Thinking with a dedicated coding model (Qwen3-Coder-30B-A3B, AWQ INT4) on one fleet GPU, so the platform can serve coding and agentic workloads. The VLM's card is freed and reassigned. FP8 KV cache is enabled fleet-wide as a performance optimization validated in the spike (#839).

---

### In Scope

- New model entry `qwen3-coder-30b-a3b` in `charts/inference/values.yaml` (vLLM, AWQ INT4, FP8 KV cache, SHA-pinned at `c58857a7f41c0920f73d1b56678640f9c02017d7`)
- Disable `qwen3-vl-4b-thinking` in `charts/inference/values.yaml` (`enabled: false`)
- Remove `qwen3-vl-4b-thinking-local` backend from `charts/ai-models/values.yaml` and set model entry to `enabled: false`
- Fleet default `kvCacheDtype: fp8_e4m3fn` in `defaults:` section of `charts/inference/values.yaml`
- 4 fail-fast validation guards in `charts/inference/templates/_helpers.tpl` (wrong engine, invalid value)
- `--kv-cache-dtype` flag in vLLM engine args, `--cache-type-k/--cache-type-v` in llama.cpp engine args
- CI fixture update in `charts/inference-server/ci/vllm-values.yaml` for `--kv-cache-dtype fp8_e4m3fn`

### Out of Scope

- Federating the coding model into the gateway (requires load gate first — ADR-0101)
- Benchmark report on inference-ops (#837)
- SGLang / TensorRT-LLM engine profiles (evaluated in spike #839, deferred)
- HumanEval/MBPP pass-rate measurement (post-merge load gate)
- Disabling or repointing the `adorsys-tiny` alias (references the removed backend — must be handled separately)

---

## 4. Verification

I verified this change by:

- [x] Running automated tests
- [ ] Running manual tests
- [x] Checking logs
- [ ] Checking metrics
- [x] Testing error cases

Commands run:

```bash
# Lint the inference chart
helm dep build charts/inference
helm lint charts/inference --strict

# Verify render: new model present, old model absent
helm template chk charts/inference | grep -c "qwen3-coder-30b-a3b"
helm template chk charts/inference | grep -c "qwen3-vl-4b-thinking"

# Lint the inference-server leaf chart with CI fixture
helm dep build charts/inference-server
helm lint charts/inference-server -f charts/inference-server/ci/vllm-values.yaml --strict

# Verify catalog consistency (served ↔ federated)
./tools/check-model-catalogs.sh
```

Results:

```text
$ helm lint charts/inference --strict
==> Linting charts/inference
[INFO] Chart.yaml: icon is recommended
1 chart(s) linted, 0 chart(s) failed

$ helm template chk charts/inference | grep -c "qwen3-coder-30b-a3b"
11

$ helm template chk charts/inference | grep -c "qwen3-vl-4b-thinking"
0

$ helm lint charts/inference-server -f charts/inference-server/ci/vllm-values.yaml --strict
1 chart(s) linted, 0 chart(s) failed

$ ./tools/check-model-catalogs.sh
check-model-catalogs: OK (2 served on the GPU fleet, 1 federated cluster-local)
  note: 'qwen3-coder-30b-a3b' is served but not federated — no user can route to it yet.
```

Negative tests (fail-fast guards):

```bash
# Invalid kvCacheDtype value → render fails
helm template chk charts/inference --set models.qwen3-coder-30b-a3b.serving.kvCacheDtype=invalid
# → Error: model qwen3-coder-30b-a3b: serving.kvCacheDtype must be one of [auto fp8 fp8_e5m2 fp8_e4m3fn]

# kvCacheDtype on llama.cpp engine → render fails
# → Error: model ...: serving.kvCacheDtype is a vLLM knob

# kvCacheType on vLLM engine → render fails
# → Error: model ...: serving.kvCacheType is a llama.cpp knob
```

---

## 5. Screenshots / Evidence

* Rendered engine args for the new model:
  ```
  vllm serve /models/Qwen3-Coder-30B-A3B-Instruct-AWQ \
    --served-model-name qwen3-coder-30b-a3b \
    --max-model-len 8192 --max-num-seqs 2 \
    --gpu-memory-utilization 0.9 \
    --quantization awq --dtype float16 \
    --kv-cache-dtype fp8_e4m3fn \
    --enable-auto-tool-choice --tool-call-parser hermes
  ```
* Model source: `QuantTrio/Qwen3-Coder-30B-A3B-Instruct-AWQ` @ `c58857a7f41c0920f73d1b56678640f9c02017d7`
* Weights: ~16.8 GB on disk (6 safetensors shards), PVC 35 GiB, seedMaxWorkers 2, seedMemoryLimit 10Gi
* VRAM estimate: ~15.6 GiB weights + 1.5 GiB overhead + 0.3 GiB KV FP8 (8K context) = **~17.4 GiB / 19.99 GiB** — fits with 2.6 GiB headroom
* Scheduling: `runtimeClassName: nvidia`, `nodeSelector: nvidia.com/gpu.present: "true"`, `nvidia.com/gpu: 1`, `storageClassName: longhorn`, probes on `/health`

---

## 6. Risk Assessment

Risk level:

* [ ] Low
* [x] Medium
* [ ] High

Potential risks:

* **PVC sizing**: 35 GiB for 16.8 GB on disk — staging in `.cache` peaks at ~2× (same mechanism that caused the qwen3-vl-4b-thinking incident #807). Mitigated by sizing at 35 GiB (~2× + margin) and capping seedMaxWorkers at 2
* **MoE architecture on vLLM**: `qwen3_moe` is the first MoE model on this fleet. vLLM supports it natively, but the AWQ Marlin MoE path is unmeasured on this specific card (RTX 4000 SFF Ada)
* **FP8 KV cache quality**: fleet default `fp8_e4m3fn` is unmeasured on this model — if the quality gate fails, override with `kvCacheDtype: auto` (16-bit) per model
* **adorsys-tiny alias**: references the removed `qwen3-vl-4b-thinking-local` backend — must be disabled or repointed in a follow-up PR

Mitigation:

- Conservative contextSize (8192) — raise after load gate measures actual headroom
- `gpuMemoryUtilization: 0.90` with ~3.4 GiB headroom after weights
- SHA-pinned weights (`c58857a7`) — reproducible seed, byte-identical on re-seed
- Rollback = `git revert` the merge commit (weights survive via `reclaimPolicy: Retain` on the Longhorn PV)
- Per-model `kvCacheDtype: auto` exists as instant rollback if FP8 quality fails

---

## 7. AI Usage Declaration

AI was used for:

* [x] Understanding existing code
* [x] Generating code
* [ ] Refactoring
* [ ] Generating tests
* [x] Drafting documentation
* [ ] Reviewing the diff
* [ ] Not used

Human verification:

---

## 8. Reviewer Focus

Please focus your review on:

* [x] Correctness — does the model entry produce the right vLLM command?
* [ ] Architecture
* [ ] Security
* [ ] Performance
* [ ] Tests
* [x] Maintainability — is the FP8 KV cache default clearly documented?
* [x] Product intent — does this replace Qwen-VL correctly?
* [x] Edge cases — what happens if the MoE AWQ path fails on this card?


Koufan-De-King (changes requested):
Reviewed by checking out the branch, rendering both charts, verifying every numeric claim against the live Hugging Face API, and running the CI-equivalent lint loop from `CLAUDE.md`.

**Verdict: request changes — 2 blockers.**

---

## 🔴 1. `adorsys-tiny` will break manifest generation on merge

The PR deletes the `qwen3-vl-4b-thinking-local` backend, but `adorsys-tiny` (`charts/ai-models/values.yaml:1976`) still references it and has **no `enabled:` key**, so it still generates a child Application. Rendering that child from this branch:

```
$ helm template adorsys-tiny charts/ai-model -f <valuesYaml extracted from the rendered ApplicationSet>
Error: execution error at (ai-model/templates/aigatewayroute.yaml:127:12):
Model 'adorsys-tiny' references backend 'qwen3-vl-4b-thinking-local' not present in backendsInventory
```

That is not a degraded alias — it is a **hard render failure**, so `aii-adorsys-tiny` goes `ComparisonError`/`Unknown` and stops syncing.

**Why every gate stayed green** (this is a real hole in the gates, not author carelessness):

- `helm template charts/ai-models` renders only the ApplicationSet — the child values are an opaque `valuesYaml` string, so the dangling ref is never evaluated.
- CI lints `charts/ai-model` against its `ci/*-values.yaml` fixtures, never against the generated per-model values.
- `tools/check-model-catalogs.sh` explicitly punts on this direction. Its own header says: *"The complementary case ('model enabled in ai-models but no backend declared') is already a hard render failure in charts/ai-model, via its `backendsInventory` guard — no need to duplicate it here."* It delegates to a render that CI does not run.

§3 lists this as an out-of-scope follow-up, but **this merge is what breaks it**. It has to land in this PR: either set `enabled: false` on `adorsys-tiny`, or repoint it at the new backend.

## 🔴 2. `toolCallParser: hermes` is the wrong parser for Qwen3-Coder

vLLM's tool-calling documentation lists `Qwen/Qwen3-Coder-30B-A3B-Instruct` under **`--tool-call-parser qwen3_xml`**. `hermes` is documented for `Qwen2.5-*` / `QwQ-32B`, whose chat templates already carry Hermes-style tool use. Qwen3-Coder emits an XML function-call format instead.

Strong corroboration: the pinned quant repo **ships its own `qwen3coder_tool_parser.py`** plugin file. That file exists precisely because the stock parsers do not fit this model.

The failure mode is quiet — tool calls come back as plain text inside `content` rather than as `tool_calls`. That is the entire value proposition of a coding model, and it is reviewer-focus item #1 in this PR. The repo's `hermes` convention (7 charts + `docs/models/qwen3-4b.md:130`) is correct for the Qwen3 *base* family; **Coder is the exception.**

Refs: https://docs.vllm.ai/en/stable/features/tool_calling/ · https://huggingface.co/QuantTrio/Qwen3-Coder-30B-A3B-Instruct-AWQ

---

## 🟠 3. The VRAM headroom number is wrong, and the real margin is tight

`--gpu-memory-utilization 0.9` × 19.99 GiB = **17.99 GiB total budget**. Weights are 15.66 GiB, leaving **~2.3 GiB** for KV cache + activations + CUDA graphs.

- The values comment says *"We have ~3.4 GiB headroom after weights"* — that matches neither `0.9 × 19.99 − 15.66 = 2.33` nor `19.99 − 15.66 = 4.33`.
- The PR body's *"~2.6 GiB headroom"* measures against card total, which is a different quantity: `gpu-memory-utilization` is a cap, and vLLM will size the KV cache to fill it.

CUDA graphs are enabled (`--enforce-eager` deliberately not set). That default was reasoned in `_helpers.tpl` for a 4B model on a 20 GiB card, not for a 48-layer / 128-expert MoE where capture memory is materially larger. The most likely first failure is vLLM aborting at startup with no memory for KV blocks.

Suggest: fix the comment, and pre-declare `--enforce-eager` via `extraArgs` as the first lever in §6 rather than discovering it during the load gate.

## 🟠 4. Two lossy layers stacked, neither in §6 Risk Assessment

The pinned quant's own model card leads with:

> ❗❗Reminder❗❗ This model suffers from significant loss under 4-bit quantization, please use with caution.

On top of that, this PR adds FP8 KV cache with uncalibrated scales. Both may well be fine in practice — but for a model whose entire value is *correctness*, the risk section should say this **before** the load gate, not after.

## 🟠 5. A fleet-wide default with no ADR

Blast radius today is smaller than the framing suggests — I verified that on this branch the **only enabled vLLM model is the new one** (`qwen3-vl-4b-thinking`, `qwen3-8b-fast` and `openmythos-27b` are all disabled; `z-image-turbo` is LocalAI and correctly unaffected). So "fleet default" is presently "the default for one model".

The cost is deferred rather than absent: it silently changes the baseline of any model re-enabled later, which the `qwen3-vl-4b-thinking` comment honestly acknowledges. Per `CLAUDE.md`'s ADR rule (decisions with non-obvious consequences or a locked-in contract get an ADR) and ADR-0101's measured-not-assumed discipline, either split the knob into its own PR + ADR, or write the ADR here.

## 🟠 6. The missing fifth guard: LMCache + fp8

`charts/inference/values.yaml` states:

> ⚠️ fp8 KV cache + LMCache is UNVERIFIED on this fleet … Verify the combination before enabling it — do not assume.

…and then makes fp8 the default, so flipping `lmcache: { enabled: true }` on any model silently produces exactly that unverified combination. Four guards were written for engine/value mismatches; this is the natural fifth and it is ~3 lines.

---

## Minor

- **`contextSize: 8192`** against the model's native `max_position_embeddings: 262144`. For agentic coding through opencode, system prompt + tool schemas + file context can exceed 8K on the first turn — the load gate may end up measuring a configuration no user would accept. Deliberate and documented; flagging the product-intent risk.
- **`dtype: float16`** — the model is `torch_dtype: bfloat16`. The repo's AWQ convention is float16 and that is defensible, but the card sets neither, so a one-line "why" would help.
- **Stray re-indent** of the `memoryLimit` comment at `charts/inference/values.yaml:733` (`# (decode/resize) runs on host CPU/RAM…` gained a space) — pure diff noise.
- **`inference.serverArgs` docstring** — the `Input: dict …` line was not updated for the new `kvCacheDtype` key (nor the pre-existing `apiKey`/`security`/`apiKeyPath`).
- **`$.kvCacheDtype`** reads from the template root while every sibling in that function reads `$s.*`. It is correct (`$` is the `include` argument), but it is the only line doing it — one clause of comment saves the next reader a lookup.
- **Documentation sweep** — 4 files changed, all charts. `CLAUDE.md`'s rule is that "update documentation" means every affected surface; at minimum `docs/patterns/self-hosted-model-serving.md` should gain the `kvCacheDtype` / `kvCacheType` knobs.

---

## Verified correct — credit where due

Every factual claim in the PR body checks out against the live HF API:

| Claim | Verified |
|---|---|
| SHA pin `c58857a7f41c0920f73d1b56678640f9c02017d7` | ✅ matches the repo's current `main` sha exactly |
| ~16.8 GB on disk, 6 safetensors shards | ✅ 16.83 GB / 15.66 GiB, 6 shards |
| MoE 128 experts, 8 active | ✅ `qwen3_moe`, `num_experts: 128`, `num_experts_per_tok: 8` |
| gate/router kept unquantized | ✅ `modules_to_not_convert: ['.mlp.gate']` |
| Apache-2.0 | ✅ |
| ~0.3 GiB KV at fp8 / 8K context | ✅ 48 KiB/token → 0.38 GiB |

Also confirmed by render/inspection:

- PVC 35Gi is correctly sized against the ~2× staging peak (33.6 GB), with `seedMaxWorkers: 2` — the #807 lesson properly applied.
- The four fail-fast guards are correct, and ordered so engine-mismatch fires before value-validation.
- The fleet default does **not** leak into the llama.cpp or LocalAI argument lists (verified by rendering both).
- The repo-wide CI-equivalent lint loop passes clean after `helm dep build`; all three `charts/inference-server/ci/*-values.yaml` fixtures lint **and** render OK.

The engineering underneath this is careful — the blockers are both at the seams between charts, which is exactly where the current gates do not look.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

rustnew (commented):
Thank you for the review — the two blockers and the orange items have been addressed. Point‑by‑point responses:

**🔴 Blocker 1 — `adorsys-tiny` / `qwen3-vl-4b-thinking-local`**
Fixed. `adorsys-tiny` has been set to `enabled: false` (`charts/ai-models/values.yaml`) and the `qwen3-vl-4b-thinking-local` backend has been removed. Verified by rendering: `helm template charts/ai-models` no longer generates a child Application for these models (the `applicationset.yaml` template excludes models with `enabled: false`), so the hard‑fail on `backendsInventory` in `aigatewayroute.yaml` is no longer reachable.

**🔴 Blocker 2 — `toolCallParser: hermes`**
Fixed → switched to `qwen3_xml`. Qwen3‑Coder emits XML‑formatted function calls (the quantized repo actually provides its own `qwen3coder_tool_parser.py`). `hermes` is correct for the *base* Qwen3 family, but not for Coder.

**🟠 3 — VRAM headroom + `--enforce-eager`**
The comment has been corrected (0.90 × 19.99 = 17.99 GiB budget, ~15.66 GiB weights → ~2.33 GiB, not ~3.4). `--enforce-eager` is pre‑declared via `serving.extraArgs` — CUDA graph capture is significantly larger on this 128‑expert MoE, and a 2.33 GiB margin is too tight to rely on. To be removed once the load gate has been measured.

**🟠 4 — Two stacked lossy layers**
Added to the model's risk section: AWQ INT4 (the quant card itself warns about the loss) + FP8 KV cache with uncalibrated scales. The load gate must measure **quality**, not just tok/s and VRAM, before fleet‑wide rollout.

**🟠 5 — Fleet default without an ADR**
ADR‑0111 has been written (`docs/adr/0111-fp8-kv-cache-fleet-default.md`) and added to the index. It documents the `fp8_e4m3fn` default, the deferred cost (it silently changes the baseline for any model reactivated later), per‑model exceptions (`auto`, `fp8_e5m2`), and the ADR‑0101 discipline of measured‑not‑assumed.

**🟠 6 — 5th missing guard: LMCache + fp8**
Already in place (commit `87e6b6f1`): `_helpers.tpl` fail‑fast if `kvCacheDtype != auto` AND `lmcache.enabled` — the unverified combination is blocked at render time.

**Minors**
- `contextSize: 8192`: kept conservative and documented (native is 262144); to be raised after the load gate.
- `dtype: float16`: a "why" comment has been added (AWQ convention from the repo, consistent with `qwen3-8b-fast`).
- Re‑indent `memoryLimit`: fixed.
- Docstring `serverArgs`: updated for `kvCacheDtype`.
- `$.kvCacheDtype`: added a comment explaining why it reads `$` (resolved value, not the override).
- Docs sweep: `kvCacheDtype` / `kvCacheType` documented in `docs/patterns/self-hosted-model-serving.md`.

Thank you for double‑checking the figures against the HF API — everything matches.

charts/inference-server/ci/vllm-values.yaml
```diff
@@ -104,6 +104,8 @@ modelServing:
             - awq
             - --dtype
             - float16
+            - --kv-cache-dtype
+            - fp8_e4m3fn
```
benie-joy-possi: Does this  value argument exist?

charts/inference/values.yaml
```diff
@@ -180,12 +180,36 @@ defaults:
...
+  #   • fp8_e4m3fn — 4-bit exponent / 3-bit mantissa: the accuracy pick for
+  #     inference, the production standard (TensorRT-LLM and most deployments).
+  #   • fp8_e5m2   — 5-bit exponent / 2-bit mantissa: robust to KV outliers
+  #     (e.g. vision heads); the plan B if the e4m3fn quality gate fails.
+  #   • auto       — 16-bit (bf16/fp16), the pre-quantization behaviour; how a
+  #     single model opts out without a fleet-wide change.
+  # ⚠️ vLLM-ONLY. llama.cpp has no FP8 vocabulary — its 8-bit cache type is
+  # q8_0 (int8 block quant) and it takes per-model `serving.kvCacheType`; the
+  # LocalAI diffusion backend has no KV cache to quantize.
+  kvCacheDtype: fp8_e4m3fn
```
benie-joy-possi: is it not just `fp8_e4m3`?

charts/ai-models/values.yaml
```diff
@@ -594,20 +594,6 @@ backends:
...
-      hostname: qwen3-vl-4b-thinking.inference.svc.cluster.local
-      port: 8080
-    securityType: APIKey
-    resourceName: qwen3-vl-4b-thinking-local-svc
-    secretRef:
-      name: qwen3-vl-4b-thinking-api-key
-    externalSecret:
-      key: ai/camer/digital/prod/env
-      property: vllm_local_api_key
-
```
benie-joy-possi, line 616: If you removing this backend  remove all the other references of it please

benie-joy-possi (changes requested):

charts/inference-server/ci/vllm-values.yaml
rustnew: yes is   precision on kvcache

charts/inference/values.yaml
rustnew: yes is   type quantization on  kvcache

charts/ai-models/values.yaml
Dericko681: I think you should keep the backend and just disable the model

PR #913, opened 2026-08-04
docs(coder): record declarative coder template management (ADR-0124)

benie-joy-possi:
## 1. Summary

This PR changes:

- Adds **ADR-0124** (`docs/adr/0124-declarative-coder-template-management.md`) recommending declarative Coder template management via the coder/coderd Terraform provider and retiring Coder templates push in production
- Adds the consolidated implementation pattern `docs/patterns/coder-declarative-templates.md`
- Indexes the new ADR and pattern in `docs/adr/README.md`, `docs/patterns/README.md`, and `docs/README.md`

It solves:

- Ticket #834

> **Note:** renumbered from ADR-0120 → **ADR-0122 → ADR-0124** (0120 was taken by `0120-provider-billing-exporter.md`; 0122 was later taken by `0122-self-hosted-code-interpreter.md`). All review feedback on the original ADR has been applied — see the summary comment on this PR.

---

## 2. Intent

The intent of this PR is:

> Resolve how coder workspace templates are delivered so workspaces are reproducible from git rather than dependent on a manual coder templates push to a live server.

---

### In Scope

- ADR-0124
- Consolidated research/implementation pattern doc

### Out of Scope

- Building the actual workspace templates - tracked as a follow-up per the pattern doc/ADR

---

## 4. Verification

I verified this change by:

- [ ] Running automated tests
- [ ] Running manual tests
- [x] Checking upstream sources
- [ ] Checking metrics
- [ ] Testing error cases

Commands run:

```bash
N/A
```

Results:

```text
N/A
```

---

## 5. Screenshots / Evidence

Add evidence here:

* Docs: `docs/adr/0124-declarative-coder-template-management.md`, `docs/patterns/coder-declarative-templates.md`

---

## 6. Risk Assessment

Risk level:

* [x] Low
* [ ] Medium
* [ ] High

Potential risks:

* Documentation-only change; no runtime/infra impact

Mitigation:

* Facts pinned to the specific provider release and verified against upstream sources

---

## 7. AI Usage Declaration

AI was used for:

* [ ] Generating code
* [ ] Refactoring
* [ ] Generating tests
* [x] Drafting documentation
* [ ] Reviewing the diff
* [ ] Not used

Human verification:

* [ ] I checked generated code manually

---

## 8. Reviewer Focus

Please focus your review on:

* [x] Correctness
* [ ] Architecture
* [ ] Security
* [ ] Performance
* [ ] Tests
* [x] Maintainability
* [ ] Product intent
* [x] Edge cases


docs/adr/0120-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
+# ADR-0120: Declarative Coder template management via the `coder/coderd` Terraform provider
```
stephane-segning: 🔴 **ADR-0120 is already taken — this needs to be 0122.**

`main` currently has:

```
docs/adr/0119-weekly-sub-budget-anti-front-loading.md
docs/adr/0120-provider-billing-exporter.md          ← Accepted, 2026-08-05
docs/adr/0121-coder-workspace-url-exposure-strategy.md
```

This branch is **28 commits behind `main`**, so the collision doesn't show up in the diff — the `docs/adr/README.md` hunk still has `0118` as the last row and will insert a second `0120` row on merge.

The repo already carries one duplicate pair (`0118-fp8-kv-cache-fleet-default.md` vs `0118-grafana-pdf-export.md`), which is a decent argument for not letting a second one land.

Needs: rebase on `main`, rename the file to `0122-…`, update the `# ADR-0122:` header, re-place the index row after `0121`, and update the two cross-links from `docs/patterns/coder-declarative-templates.md`.

Since ADR-0121 is *also* a Coder ADR (workspace URL exposure), a rebase would be worth doing regardless — the two should probably reference each other.

docs/adr/0120-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Context
+
+Coder workspace templates are currently pushed to the live Coder deployment
+imperatively with `coder templates push` (from ticket #834's parent epic #822,
+which wants workspaces reproducible from git, no manual CLI push required for a
+fresh cluster). `coder templates push` is stateless: no version control of the
+deployment action, no rollback, no drift detection, and it does not follow
+symlinks ([coder/coder#5677](https://github.com/coder/coder/issues/5677)) — a
+known trap for git-backed template directories.
```
stephane-segning: 🔴 **This citation doesn't support the claim.** I pulled the issue:

```
Title:  Support terraform resources outside the given template directory
Opened: 2023-01-11   Closed: 2023-01-16 (completed)
```

Its body asks for the ability to reference shared Terraform modules living outside a template's directory, so near-identical templates don't have to copy-paste common resources. **"symlink" appears nowhere in the title or body**, and the issue has been closed for three and a half years.

This matters more than a bad link because it's load-bearing: it's one of only two grounds given for rejecting the CLI at L74-76, and it's repeated in the pattern doc twice (`pattern.md:37`, `pattern.md:180-181`).

Either find the real source for the symlink behaviour (if `push`'s tar step genuinely doesn't follow them, that's worth citing properly) or drop the claim. The other reasons for the decision — stateless, no drift detection, no rollback — stand on their own and don't need it.

docs/adr/0124-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+symlinks ([coder/coder#5677](https://github.com/coder/coder/issues/5677)) — a
+known trap for git-backed template directories.
+
+Coder ships a first-party Terraform provider, `coder/coderd` (registry
+`coder/coderd`, current release `v0.0.22`, requires Coder `>= 2.10.1`; our
+deployment runs v2.34.6, current upstream v2.34.7). Its `coderd_template`
+resource manages the full template lifecycle — create, version from a
+directory, activate/deactivate, ACLs (Enterprise), TTL/dormancy, deprecation —
+declaratively. Research (verified against the provider's current documentation,
+Aug 2026) found no parity gap that blocks adopting it over `push`.
```
stephane-segning, line 25: Two of these four numbers are off. Checked against the GitHub release feeds just now:

| Claim | Actual |
|---|---|
| `coder/coderd` `v0.0.22` | ✅ correct (published 2026-07-13) |
| requires Coder `>= 2.10.1` | ✅ correct |
| "current upstream v2.34.7" | ❌ **v2.35.3**, published 2026-07-27 |
| "our deployment runs v2.34.6" | ❌ repo pins the chart at **`2.33.7`** (`charts/coder/values.yaml:74`) |

The v2.35.3 release predates this ADR's 2026-08-03 date, so the "checked online, Aug 2026" framing was already stale on arrival.

Neither error changes the decision — both versions clear `>= 2.10.1` comfortably — but this is the kind of table people quote later without re-checking. If the live deployment genuinely differs from the chart pin, that's worth saying explicitly rather than picking one.

docs/adr/0120-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+  resources, `versions[].directory` pointing at the source, remote state on
+  **Hetzner Object Storage** via the S3-compatible backend, Git-SHA-encoded
+  version names). The provider content-hashes each template directory at plan
+  time and creates a new Coder version when the hash changes; the SHA-in-name
+  gives each new content-hash a distinct, traceable version name.
+- Deployment flows through a GitOps pipeline (PR validates via
+  `terraform plan`; merge to `main` runs `terraform apply`; optional scheduled
+  reconciliation detects drift).
+- New/fresh Coder clusters reproduce templates from Git + Terraform state + a
+  token — no manual `coder templates push`.
```
stephane-segning: **The fresh-cluster claim doesn't survive contact with a fresh cluster.** (Same text at `pattern.md:142-143`.)

"Reproduce from Git + Terraform state + a token" works for *disaster recovery of the same deployment*. But for a genuinely new Coder deployment — the case the parent epic #822 actually cares about — the existing state is not just unhelpful, it's wrong: it holds template and version UUIDs belonging to the dead deployment. `apply` against a new Coder either 404s on refresh or tries to update resources that don't exist; you'd need `terraform state rm` (or a separate workspace/state key) first.

The honest version is probably: *fresh cluster = Git + a token + **empty** state* — which is a stronger claim anyway, since it means the state isn't a DR dependency at all. Worth one sentence, because "no manual push for a fresh cluster" is the headline benefit and it's the one step not worked through.

Related and also unaddressed: the pipeline needs a **Coder admin token** and **S3 credentials** as GitHub Actions secrets. The repo's secret convention is ESO → cluster (`ssegning-aws`), which doesn't reach Actions — so this introduces a new secret surface with no stated owner, rotation, or provisioning path. And per CLAUDE.md the follow-up workflow will need `runs-on: adorsys-gis-runner`, since GitHub-hosted runners are billing-blocked org-wide.

docs/adr/0120-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+
+**Positive**
+- Single source of truth via Terraform remote state; deterministic,
+  reproducible template delivery from git.
+- Drift detection (`terraform plan`) and rollback (revert git / restore state)
+  where the CLI has none.
+- Full audit trail through PR review; aligns with the platform's existing IaC/
+  GitOps conventions.
+
+- **Negative**
```
stephane-segning: Stray leading `- ` — this renders as a bullet *inside* the Positive list rather than as the Negative heading. Should be `**Negative**` to match `**Positive**` (L47) and `**Neutral / follow-ups**` (L66).

docs/adr/0124-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+- **CLI-only (`coder templates push`)** — rejected: stateless, not
+  GitOps-compliant, no drift detection or rollback, symlink limitation
+  (coder/coder#5677).
+- **REST API + custom scripts** — rejected: still imperative, DIY automation
+  with no state management, reinvents what the provider already does.
+- **Helm values / Argo CD App as the delivery mechanism** — rejected: Coder
+  does not expose templates as Helm values, and Argo CD is for cluster
+  resources, not a stateful Coder template store; GitHub Actions +
+  `terraform apply` is the declared source of truth (Argo CD only for drift
+  visibility).
```
stephane-segning, line 93: The CLI rejection rests on four reasons, one of which ("symlink limitation") is the misread issue from L13 — see that thread. The other three hold, so this just needs the bad one removed rather than a rewrite.

The third bullet is the interesting one and I think it's argued correctly: this repo's delivery model is ArgoCD continuous delivery (ADR-0055), so introducing a **second** delivery mechanism — GitHub Actions running `terraform apply` — is a real architectural cost, and it's good that the ADR names it rather than glossing over it. Worth being slightly more explicit that this is the repo's *first* Terraform surface and its first Actions-driven apply, so a reader understands the precedent being set.

stephane-segning (commented):
## Review — ADR-0120 + declarative Coder templates pattern

The decision itself is sound and I'd land it: `coderd_template` really does cover the lifecycle, the parity table is the right way to frame the choice, and the content-hash-vs-version-name explanation (`pattern.md:123-132`) is genuinely useful — that's the non-obvious bit people get wrong, and it's explained well. All three indexes were updated per the docs rule, and the Alternatives are real alternatives with real reasons.

But this is a research spike whose entire value is that its facts are right, and I checked them against upstream. Two are wrong in ways that matter, and one is a merge blocker.

### 🔴 The ADR number is already taken
`main` has **`docs/adr/0120-provider-billing-exporter.md`** (Accepted, 2026-08-05) and `0121-coder-workspace-url-exposure-strategy.md`. This branch is **28 commits behind `main`**, which is why the collision isn't visible in the diff — the `docs/adr/README.md` hunk still shows `0118` as the last row, so merging adds a *second* `0120` row to the index. Next free number is **0122**. (The repo already carries one duplicate — `0118-fp8-kv-cache-fleet-default.md` vs `0118-grafana-pdf-export.md` — which is a good argument for not adding another.) Needs a rebase + renumber of the file, the header, the index row, and the two cross-links from the pattern doc.

### 🔴 `coder/coder#5677` does not say what the ADR says it says
Cited three times as the live "`push` does not follow symlinks" limitation, and it is one of only two stated grounds for rejecting the CLI in Alternatives. Actual issue:

> **Title:** Support terraform resources outside the given template directory
> **Opened** 2023-01-11 · **Closed as completed** 2023-01-16

The word "symlink" appears nowhere in the title or body — it's a 2023 feature request about *referencing files outside the template directory*, and it was closed three and a half years ago. Either find the real citation for the symlink behaviour or drop the claim; as it stands a load-bearing argument rests on a misread, closed issue.

### 🟠 `versions[].description` is not a valid attribute
The canonical implementation example (`pattern.md:113`) uses `description = "Managed version"`. The provider schema is:

> **Required:** `directory` · **Optional:** `active`, `message`, `name`, `provisioner_tags`, `tf_vars` · **Read-only:** `directory_hash`, `id`

So that example fails with `Unsupported argument`. The attribute is **`message`** — and notably #928's own `templates.tf` uses `message` correctly, so the doc contradicts the implementation it exists to guide. Worth fixing precisely because someone will copy this block.

### 🟠 Version numbers in the "Verified facts" table are stale
- ✅ `coder/coderd v0.0.22` — correct (released 2026-07-13).
- ✅ Requires Coder `>= 2.10.1` — correct.
- ❌ "Current Coder release `v2.34.7`" — latest is **v2.35.3**, released 2026-07-27, i.e. *before* this ADR's 2026-08-03 date.
- ❌ "we run `v2.34.6`" — this repo pins the Coder chart at **`2.33.7`** (`charts/coder/values.yaml:74`).

Neither changes the conclusion (both are comfortably ≥ 2.10.1), but a table headed "Verified facts (checked online, Aug 2026)" is exactly the thing future readers will trust without re-checking.

### 🟠 The recommended state backend has no locking — and locking is why an alternative was rejected
Git-backed state is rejected at `pattern.md:159-160` because it "has no locking", but the proposed S3 block has no `dynamodb_table` (Hetzner has no DynamoDB) and no `use_lockfile = true`. As written, the recommendation has the same gap it rejects. Details inline at `pattern.md:84`, along with two deprecated arguments in the same block.

### Also worth addressing
- **The fresh-cluster DR claim doesn't work as stated** (`adr.md:42`, `pattern.md:142`) — for a genuinely new Coder deployment the old state is *actively wrong*.
- **Pipeline credentials are unaddressed** — the flow needs a Coder admin token + S3 keys as GitHub Actions secrets, a surface the repo's ESO→cluster convention doesn't cover, with no stated owner or rotation story. The follow-up workflow will also need `runs-on: adorsys-gis-runner` (GitHub-hosted runners are billing-blocked org-wide).
- `- **Negative**` at `adr.md:55` renders as a bullet inside the Positive list.

### Verdict
Accept the decision, fix the citations. Concretely, before merge: rebase + renumber to **0122**; fix or drop the #5677 claim; `description` → `message`; correct the two version numbers; add `use_lockfile` + the un-deprecated S3 arguments.

Worth noting for sequencing: #928 already implements this ADR's layout, so whichever lands first should carry the `terraform/coder/` scaffold (`main.tf`, `variables.tf`, backend, workflow) that both PRs currently assume and neither provides.

---
*Review by Claude Opus 5 at @stephane-segning's request. Every external claim above was re-checked against upstream (GitHub API for the issue and both release feeds, the provider's own `docs/resources/template.md` for the schema) rather than taken from the doc.*

docs/adr/0124-declarative-coder-template-management.md
```diff
@@ -0,0 +1,98 @@
...
+
+## Related
+
+- Docs: [`docs/patterns/coder-declarative-templates.md`](../patterns/coder-declarative-templates.md)
+  (the *how*: verified facts, parity table, repo layout, implementation pattern)
+- Tickets: [#834](https://github.com/ADORSYS-GIS/ai-helm/issues/834)
+  (this spike), epic #822
```
stephane-segning, line 103: Two ADRs missing from Related that a reader will want:

- **ADR-0083** (`0083-coder-re-introduction.md`) — the parent decision that put Coder in the platform at all, and the reason `charts/coder` exists. This ADR is downstream of it.
- **ADR-0121** (`0121-coder-workspace-url-exposure-strategy.md`) — already on `main`, also about Coder workspaces. Since renumbering to 0122 puts this ADR directly after it, a cross-reference in both directions would help.

(ADR-0027 is also arguably relevant as the *removal* that 0083 superseded, but that's optional.)

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+| `coder templates push` CLI | still exists (imperative path) |
+
+References:
+- Provider registry: https://registry.terraform.io/providers/coder/coderd/latest
+- Provider source: https://github.com/coder/terraform-provider-coderd
+- `coderd_template` resource docs: https://registry.terraform.io/providers/coder/coderd/latest/docs/resources/template
+- `coder templates push` CLI: https://coder.com/docs/reference/cli/templates_push
+- Templates as Terraform modules: https://coder.com/docs/admin/templates/extending-templates/modules
+- Provider GitOps announcement: https://coder.com/blog/release-recap-2-15-0
+- Known limitation (push does not follow symlinks): https://github.com/coder/coder/issues/5677
```
stephane-segning: Corrections to this table and reference list — see the ADR L17 thread for sources:

- **L26** — "Current Coder release `v2.34.7` (we run `v2.34.6`)" → latest is **v2.35.3** (2026-07-27); the repo pins `2.33.7` (`charts/coder/values.yaml:74`).
- **L37** — "Known limitation (push does not follow symlinks)" → coder/coder#5677 is *"Support terraform resources outside the given template directory"*, closed 2023-01-16, no mention of symlinks. Same fix needed at L180-181.

L24-25, L27-28 all check out. Given the table's whole purpose is to be the trustworthy snapshot, it's worth adding the date each row was verified rather than one blanket "Aug 2026" in the heading — that way a stale row is visible as stale.

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+  backend "s3" {
+    bucket                      = "ssegning-k8s-state"
+    key                         = "coder/templates/terraform.tfstate"
+    region                      = "us-east-1"              # Ceph-RGW accepts this
+    endpoint                    = "https://nbg1.your-objectstorage.com"
+    skip_credentials_validation = true
+    skip_requesting_account_id  = true
+    skip_metadata_api_check     = true
+    force_path_style            = true
+  }
```
stephane-segning, line 100: Three issues in this block — worth getting right since it's meant to be copy-pasteable.

**1. Two deprecated arguments.** Deprecated since Terraform 1.6.0 (the AWS SDK Go v2 migration):
- `endpoint` → `endpoints.s3`
- `force_path_style` → `use_path_style`

**2. No state locking.** There's no `dynamodb_table` (Hetzner has no DynamoDB) and no `use_lockfile`. So two concurrent applies — say a merge-to-main deploy racing the "optional daily reconcile" workflow this doc proposes at L139 — can corrupt state. Terraform ≥1.10 supports S3-native locking via `use_lockfile = true`, which needs conditional-write (`If-None-Match`) support; worth verifying Ceph-RGW honours it on Hetzner before relying on it.

**3. Probably needs `skip_s3_checksum = true`.** AWS SDK Go v2's default integrity checksums break uploads against several S3-compatible stores. Flagging as "verify at implementation time" rather than as fact — I haven't tested it against this bucket.

```hcl
backend "s3" {
  bucket = "ssegning-k8s-state"
  key    = "coder/templates/terraform.tfstate"
  region = "us-east-1"

  endpoints                   = { s3 = "https://nbg1.your-objectstorage.com" }
  use_path_style              = true
  use_lockfile                = true   # verify RGW conditional-write support
  skip_s3_checksum            = true   # verify
  skip_credentials_validation = true
  skip_requesting_account_id  = true
  skip_metadata_api_check     = true
}
```

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+  display_name    = "Ubuntu Development"
+  description     = "Full-featured Ubuntu development environment"
+  organization_id = var.organization_id
+
+  versions = [{
+    name        = "v-${var.git_commit_sha}"
+    description = "Managed version"
+    directory   = "${path.root}/../../.coder/templates/ubuntu"
+    active      = true
+  }]
```
stephane-segning, line 124: 🟠 **`description` is not a valid attribute inside `versions` — it's `message`.** Per the provider's own resource docs:

> **Required:** `directory`
> **Optional:** `active`, `message`, `name`, `provisioner_tags`, `tf_vars`
> **Read-only:** `directory_hash`, `id`

So this block fails with `Unsupported argument`. (`description` *is* valid on the resource itself at L108 — just not on a version.)

The tell that this is a real bug and not my misreading: **#928's `templates.tf` uses `message`**, so the implementation already diverges from the doc that's supposed to guide it. Worth fixing here precisely because the next person will copy this block rather than the PR.

While here — `provisioner_tags` is the one genuinely useful attribute the doc never mentions, and it's how you'd route template builds to a specific provisioner. Probably worth a row in the parity table.

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+the `ssegning-aws` store (`prod/meta/test-app`, the `s3_backup_*` properties)
+via `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`.
+
+Alternatives considered and rejected:
+- **GCS** — this repo/platform has no Google Cloud footprint; GCS was a
+  placeholder in the original spike docs.
+- **HCP Terraform (Terraform Cloud)** — managed state + runs, but a new SaaS
+  dependency + token secret for no benefit over the existing store.
+- **Git-backed (Atlantis / state committed to `ai-helm-values`)** — Terraform
+  has no official git backend; requires custom automation and has no locking.
```
stephane-segning, line 170: Git-backed state is rejected here partly because it "has no locking" — but the S3 backend recommended at L84-93 has no locking either, as written (no `dynamodb_table`, no `use_lockfile`). The reasoning and the recommendation contradict each other.

Adding `use_lockfile = true` fixes both the gap and the inconsistency; see the L84 thread. If RGW turns out not to support conditional writes, then the honest framing is "no backend available to us has locking, and here's why that's acceptable at this concurrency" — which is a fine answer, just a different one from the one currently implied.

The GCS and HCP Terraform rejections are well-reasoned and I'd keep them as-is.

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+
+Importing an already-pushed template into Terraform state (one-time):
+
+```bash
+terraform import coderd_template.ubuntu coder/ubuntu
+# then declare its versions in templates.tf to reach "no drift"
+```
```
stephane-segning, line 180: Small clarity thing: the import ID format is `<organization-name>/<template-name>`, so the `coder/` prefix here is the **organization name**, not a fixed namespace or a provider prefix. As written (`coder/ubuntu`, alongside a resource literally named `ubuntu`) it reads like boilerplate, and someone whose Coder org isn't named `coder` will copy it and get a not-found.

Suggest `terraform import coderd_template.ubuntu <org-name>/ubuntu` with a one-line note. The provider docs also accept a template UUID, which is worth mentioning as the unambiguous option.

docs/patterns/coder-declarative-templates.md
```diff
@@ -0,0 +1,187 @@
...
+  (or `tf_vars`) change (Coder identifies versions by name); encode the Git SHA
+  in the name so each new content-hash gets a unique, traceable label.
+- **Import is one-time manual** — versions must be declared after `import`.
+- **Startup metadata race** — some enterprise attributes
+  (e.g. `max_port_share_level`) may need a second apply after first creation.
+- **Global/workspace server settings** — not in the provider; manage via UI/API
+  or Helm (server-level admin is tracked separately).
+- **Multi-org** — provider manages one org per config; use separate workspaces/configs.
+- **`templates push` symlinks** — provider path has no such tar/symlink issue,
+  but the noted CLI limitation (coder/coder#5677) is why Git source is preferred.
```
stephane-segning: Same bad citation as `adr.md:13` — coder/coder#5677 is about referencing Terraform resources *outside* the template directory, not symlinks, and it's been closed since January 2023.

This bullet is doubly odd because it says the provider path "has no such tar/symlink issue" and then concludes that's "why Git source is preferred" — but Git source is where template files live under *both* delivery paths, so the symlink behaviour wouldn't distinguish them even if the claim were accurate. Cleanest fix is to delete the bullet; the other five limitations in this list are real and well-stated.

PR #928, opened 2026-08-05
feat(coder): per-workspace ServiceAccount auth for OpenCode in Coder workspace (#835)

### LINKED ISSUE #835: [Ticket]: Complete configuration of Coder OpenCode agent
> Upgraded to the AI-Helm governance **Dev Ticket** format · https://adorsys-gis.github.io/ai-governance/03-dev-ticket-template

## 1. Type
- [x] Feature / Configuration

## 2. Summary
Complete the configuration of the Coder OpenCode agent using the Coder registry module (`registry.coder.com/coder-labs/opencode/coder` v0.1.2), including AgentAPI integration for task reporting in the Coder UI, MCP server configuration, and auth.json credential management.

## 4. Source of Truth
- Parent epic: #822
- Coder OpenCode module: https://registry.coder.com/modules/coder-labs/opencode
- AgentAPI: https://github.com/coder/agentapi
- OpenCode agents: https://opencode.ai/docs/agents/
- OpenCode skills: https://opencode.ai/docs/skills/
- AgentAPI refactor discussion: https://github.com/coder/coder/discussions/21348
- Existing OpenCode work: ADR-0037, ADR-0071/0072/0073, `docs/opencode-well-known.md`

## 5. Expected Behavior / Deliverables
- [ ] Add the OpenCode module to our Coder workspace template (Terraform).
- [ ] Configure `agent_id`, `workdir`, `config_json` (with MCP servers).
- [ ] Resolve `auth.json` credential management (ESO / projected SA tokens per ADR-0037).
- [ ] Verify AgentAPI task reporting works in the Coder UI.
- [ ] Verify the OpenCode agent responds inside a Coder workspace (test a coding task).

## 6. Acceptance Criteria
- [ ] Given a Coder workspace with the OpenCode module, when the agent starts, then it is reachable and responds to coding tasks.
- [ ] Given the agent runs a task, when AgentAPI is configured, then the task appears in the Coder UI.
- [ ] Given auth.json is needed, when the workspace starts, then credentials are provisioned via ESO/SA tokens (no hand-installed secrets).

## 7. Out of Scope
- LibreChat↔Coder MCP integration (LibreChat epic #821).
- Adding new MCP servers to the OpenCode catalog (epic #549).

## 8. Technical Context
- The Coder OpenCode module (v0.1.2) integrates with AgentAPI for task reporting and supports MCP server configuration via `config_json`.
- `auth_json` is required for non-interactive authentication (found at `$HOME/.local/share/opencode/auth.json`).
- AgentAPI is being refactored to remove the dependency (coder/coder#21348) — pin module version.
- The module can run in standalone CLI mode (`report_tasks = false`, `cli_app = true`) if task reporting is not needed.

## 9. Verification Evidence
N/A — to be filled on completion.

## 10. AI Usage Declaration
AI was used for drafting this ticket and enriching technical context from the Coder registry module documentation, AgentAPI GitHub, and OpenCode docs.

## 11. Parent User Story
- Parent epic: #822

benie-joy-possi:
## 1. Summary

Adds a Coder template (`.coder/templates/opencode-task/`) that runs OpenCode in a Coder workspace and authenticates it to the **internal** AI gateway with **per-user attribution**, replacing the injected-owner-OAuth-token approach. Documented in **ADR-0130**.

This PR changes:
- **`.coder/templates/opencode-task/main.tf`** — provisions a per-workspace ServiceAccount named `coder-<sub>.<plan>` (owner's Keycloak `sub` + billing plan in the name), a workspace pod (OpenCode + an **openresty sidecar**), and wires the projected SA token (audience `core-gateway-internal`) into the sidecar only.
- **Authorino** derives both `sub` (owner) and `plan` by **pure CEL** from `kubernetesTokenReview` — no SA-label read, no K8s-API metadata call. The internal-plane `AuthConfig` change lives in `adorsys-gis/ai-helm-values` `security-policies.yaml`.
- Adds **ADR-0130** + design/flow docs + diagram.
- Removes `terraform/coder/templates.tf` (declarative-template scaffold) — out of scope for this PR.

It solves:
- #835 — Configuring the Coder + OpenCode agent end-to-end without interactive OAuth login and without placing a real user credential in the pod.

## 2. Intent

> Make OpenCode inside a Coder workspace authenticate to the AI gateway non-interactively, with per-user budget attribution (ADR-0035), all internal, and with **no real credential in the user-controlled pod**. OpenCode uses a dummy key; an openresty sidecar reads the workspace's projected SA token per request and injects it; Authorino validates via `kubernetesTokenReview` and derives identity + plan from the SA name.

### In Scope
- `opencode-task` Coder template: per-workspace SA `coder-<sub>.<plan>`, two-container pod (OpenCode + openresty sidecar), projected SA token (sidecar only), internal-CA trust, task capability via `coder_ai_task`.
- `security-policies.yaml` (ai-helm-values): workspace SA `x-account-id`/`x-billing-plan` derived from the SA name via CEL.
- ADR-0130 + docs + diagram.

### Out of Scope
- Keycloak admin changes (billing-plan mapper is an upstream config task; the `coder` client already carries `billing_plan`).
- `ai-helm-values` delivery pipeline (separate repo/PR).
- The declarative template scaffold (`terraform/coder/`) — removed here; tracked separately (ADR-0124).

## 4. Verification

- [ ] Automated tests
- [x] Manual tests
- [x] Logs
- [x] Security behavior

Commands / evidence:
```bash
terraform fmt -check main.tf
helm lint ...   # n/a (template, not chart)
# Live Authorino test (default k3s): a workspace SA token returned:
#   x-account-id: <owner sub>   x-billing-plan: pro    (from the SA name alone)
# openresty sidecar: per-request token read + header strip + /healthz 200
#   under a read-only root FS (pid + temp dirs redirected to a writable emptyDir)
```
Results:
```text
Live Authorino Check (grpcurl) with SA coder-<sub>.pro => x-account-id=<sub>, x-billing-plan=pro
openresty (read-only) healthz 200; forged client identity headers stripped
terraform validate: only the pre-existing in-cluster kubernetes provider file() errors
```

## 5. Screenshots / Evidence
- Authorino `Check` response headers (x-account-id / x-billing-plan).
- openresty read-only container healthz + header-strip logs.

## 6. Risk Assessment

Risk level:
- [ ] Low
- [x] Medium
- [ ] High

Risks / mitigations:
- Plan changes require a workspace rebuild (plan is in the SA name) — acceptable, plans change rarely.
- `sub` is parsed by split on `.` (delimiter), robust to any `sub` length and to `coder-flows`-style namespaces — verified live.
- Full end-to-end (workspace boot → real gateway → model) still to be validated on the real cluster.

## 7. AI Usage Declaration

AI was used for:

- [ ] Understanding existing code
- [x] Generating code
- [ ] Refactoring
- [ ] Generating tests
- [x] Drafting documentation
- [x] Reviewing the diff
- [ ] Not used

Human verification:
- [x] I removed unsupported AI assumptions

## 8. Reviewer Focus

Please focus on:

- [x] Correctness
- [x] Architecture
- [x] Security
- [ ] Performance
- [ ] Tests
- [x] Maintainability
- [x] Product intent
- [x] Edge cases


.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+  required_version = ">= 1.0"
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = ">= 2.13"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.23"
+    }
```
stephane-segning, line 11: **Unbounded version floors + a gitignored lock file = not reproducible.**

`>= 2.13` and `>= 2.23` admit a future v3 major, and `.gitignore` excludes `.terraform.lock.hcl` — so each template version resolves providers fresh and two builds of the same commit can differ.

That is precisely what ADR-0120 (#913) exists to prevent: "deterministic, reproducible template delivery from git." The lock file is part of the tarball the provider content-hashes and ships to Coder; it should be committed.

```hcl
coder      = { source = "coder/coder",           version = "~> 2.13" }
kubernetes = { source = "hashicorp/kubernetes",  version = "~> 2.23" }
```
and drop `.terraform.lock.hcl` from `.gitignore`.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+provider "kubernetes" {
+  # If Coder is running INSIDE k3s (same cluster), use the service account token
+  # This is the standard setup for k3s deployments
+  host                   = "https://kubernetes.default.svc"
+  cluster_ca_certificate = file("/var/run/secrets/kubernetes.io/serviceaccount/ca.crt")
+  token                  = file("/var/run/secrets/kubernetes.io/serviceaccount/token")
+
+  # If Coder is OUTSIDE k3s, comment out the above and uncomment below:
+  # config_path = "~/.kube/config"
+}
```
stephane-segning, line 25: Two things here:

1. **The commented-out alternative should go.** House rule is no dormant/commented-out parallel paths — pick the in-cluster config and delete the `config_path` branch. Same for the `# Option B (Alternative)` block at L220 and `# This is what was missing!` at L211; they read as leftover scratch.
2. **`file()` is evaluated at plan time**, so this template can't be `terraform validate`d anywhere except inside the cluster — which matches the "expected pre-existing errors" note in the PR description. That's tolerable (the provisioner does run in-cluster), but it does mean the template dir can never be covered by the PR-time `plan` job ADR-0120 specifies. Worth a comment saying so, so the next person doesn't try to "fix" the validate failure.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+  description = "Provider/server id used by @vymalo/opencode-oauth2 as the cache file stem (serverId) — must match the provider key in the remote config served by opencode_url."
+  default     = "camer-digital"
+}
+
+variable "openai_api_key" {
+  type        = string
+  description = "API key for the OpenAI-compatible endpoint"
+  sensitive   = true
+  default     = ""
+}
```
stephane-segning: **`openai_api_key` is never used for anything.** Its only consumer is the `CAMER_API_KEY` env at L251-254, which nothing in the pod reads. Please delete both — a `sensitive` variable rendered straight into a pod spec is the L259 problem waiting to recur.

Same for three more declared-but-unread variables, all of which the README documents as if they were live knobs:
- `openai_base_url` (L35) — README calls it "OpenAI-compatible API base URL"; the real base URL comes from the well-known descriptor.
- `model` (L60, default `adorsys-coder`) — README calls it "Default model name"; never wired into `opencode_config`.
- `keycloak_issuer_url` (L73) — unreferenced.

Either wire them or drop them. Documenting non-functional knobs is worse than omitting them, because the next person will set one and wonder why nothing changed.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+  # Minimal local config. The provider (camer-digital) comes from the remote
+  # config fetched from the URL; we only set runtime defaults/permissions here.
+  opencode_config = jsonencode({
+    "$schema" = "https://opencode.ai/config.json"
+    logLevel  = "DEBUG"
+    permission = {
+      edit = "allow"
+      bash = "allow"
+    }
+  })
```
stephane-segning, line 209: **`logLevel = "DEBUG"`** on a process that carries a bearer token — please use `INFO` unless there's a specific reason.

And `edit`/`bash` both `allow` deserves a second look given this template is explicitly task-capable via the API (`coder_ai_task`, L317). The combination is: an externally-submitted prompt gets unrestricted `bash` in a pod that holds both the user's gateway token (L259) and — because `automount_service_account_token` is not disabled — the `coder` namespace default ServiceAccount token. Any one of those is fine alone; all three together is a wide blast radius for a prompt you didn't write.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+    # 1. Install Node.js (Required for OpenCode)
+    if command -v apt-get &> /dev/null; then
+      sudo apt-get update
+      sudo apt-get install -y curl gnupg
+      curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -
+      sudo apt-get install -y nodejs
+    fi
+
+    # 2. Install basic tools
+    sudo apt-get install -y git build-essential
```
stephane-segning: **Step 2 escapes the guard.** Step 1 wraps the Node install in `if command -v apt-get`, but L150 is unconditional and step 3's `node -e` assumes Node exists. With `set -e` at the top, a non-apt image fails the whole startup script at L150 rather than degrading. Either extend the guard over steps 2–3 or drop it (the image is pinned to Ubuntu anyway) — right now it's half a guard.

Separately, L145 pipes `deb.nodesource.com` straight into `sudo bash` on **every** workspace start, and the image is a floating Docker Hub tag (L233, no digest). That's two unpinned remote-code paths in the boot sequence, and it re-runs on every restart because there's no PVC. Baking Node into a first-party image removes both and cuts start time materially.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+    # Terraform $$ escaping, which previously rendered a mangled "35HOME/.cache"
+    # path and wrote the seed to the wrong directory.
+
+    # (a) Wellknown auth entry -> enables remote config fetch from the URL
+    if [ -n "$${OPENCODE_OAUTH_AUTH_JSON:-}" ]; then
+      mkdir -p "$$HOME/.local/share/opencode"
+      printf '%s\n' "$${OPENCODE_OAUTH_AUTH_JSON}" > "$$HOME/.local/share/opencode/auth.json"
+      chmod 0600 "$$HOME/.local/share/opencode/auth.json"
+      echo "seeded auth.json (wellknown)"
+    fi
```
stephane-segning: 🔴 **Blocker — the `$$HOME` bug is still here.** The note at L168-170 says it was fixed, but only the node block was; this block still uses `$$HOME`.

Terraform's heredoc escape is `$${` — a bare `$$` is **not** an escape sequence, so it passes through literally. Verified with an actual apply:

```hcl
output "s" {
  value = <<-EOT
    A="$${FOO:-}"
    B="$$HOME/.local"
  EOT
}
```
```
s = <<EOT
A="${FOO:-}"      # correct — $${ IS an escape
B="$$HOME/.local" # literal $$ — NOT an escape
EOT
```

The shell then expands `$$` to the PID, so L174-176 create `./<pid>HOME/.local/share/opencode/auth.json` in the agent's cwd — a junk directory — and L177 cheerfully echoes `seeded auth.json (wellknown)`. Exactly the `35HOME` failure the comment describes.

It's masked today because `module.opencode` also receives `auth_json` (L313) and writes the real file, so the symptom is litter rather than an outage — but this block is dead code that lies in the logs.

`$HOME` needs no escaping in a Terraform heredoc (only `${` does). Either use `"$HOME/..."` or, better, delete the block and let the module own the file — see L313.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+      // not a machine flow) treats a cached token WITHOUT expiresAt as always
+      // valid (oauth/client.js#isTokenValid returns true), so it never triggers
+      // a device-code prompt or a refresh attempt. The token we inject is a
+      // long-TTL (or delivered-fresh-each-start) owner token whose real lifetime
+      // is governed by Keycloak, so we must not let the plugin re-auth based on
+      // its own expiry bookkeeping.
+      const tokenObj = { accessToken: token, tokenType: "Bearer" };
+      const data = { serverId: process.env.OPENCODE_OAUTH_SERVER_ID, models: [], rawModels: [], token: tokenObj };
+      fs.writeFileSync(file, JSON.stringify(data, null, 2) + "\n", { mode: 0o600 });
+      console.log("seeded cache " + file + " (no expiresAt; plugin treats as always-valid)");
```
stephane-segning: 🔴 **Blocker — this makes the workspace fail closed with no recovery.**

No `expiresAt` and no refresh token, against a ~5-minute Keycloak access-token TTL, means: the workspace works for ~5 minutes, then every gateway call 401s, and the plugin — having been told the token is permanently valid — will never re-auth, never refresh, never prompt. The user's only recovery is rebuilding the workspace. That's arguably a worse outcome than the device-code prompt this is designed to avoid, because at least the prompt is actionable.

The PR body acknowledges this and defers the Keycloak TTL bump as out of scope. But "authenticate non-interactively" is the entire feature — shipping the trick without the TTL change puts a template in the Coder UI that appears to work and doesn't. Worth weighing before merge:

- **`data.coder_external_auth`** — Coder's own external-auth provider handles refresh and exposes a live `access_token`. Probably the idiomatic answer, and it removes the need to lie to the plugin at all.
- **Land the Keycloak access-token TTL change first** and state it as a hard prerequisite in the template README + the ADR.
- Since we own `@vymalo/opencode-oauth2`, a first-class **injected-token mode** upstream would beat depending on `isTokenValid()` returning true for a token with no expiry — that's an implementation detail, not an API.

Also: `OPENCODE_OAUTH_NAMESPACE` (L189, `"opencode-oauth2-model-sync"`) hardcodes the plugin's private cache namespace. The org-wide descriptor pins the plugin at `@vymalo/opencode-oauth2@0.12.0` (`charts/librechat-opencode-wellknown/values.yaml:50`) and the workspace fetches its plugin list from that descriptor — so a routine bump in the well-known chart can silently break every workspace, with no test anywhere that catches it. At minimum, note the plugin version this was verified against.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+    # Option B (Alternative): If you prefer readable names, use a hash to avoid special chars
+    # name      = "coder-${substr(md5("${data.coder_workspace_owner.me.name}-${data.coder_workspace.me.name}"), 0, 16)}"
+
+    namespace = var.namespace
+    labels = {
+      "app" = "coder-workspace"
+    }
+  }
+  spec {
+    restart_policy = "Always" # Changed from Never to keep workspace alive
```
stephane-segning: `restart_policy = "Always"` on a bare Pod means a startup-script failure produces a restart loop that re-runs the full apt + nodesource install on every cycle. Coder's own Kubernetes templates use `Never`, so a broken workspace surfaces as stopped rather than as an infinite loop.

The comment ("Changed from Never to keep workspace alive") suggests a different problem was being worked around — what was it? If the pod was exiting, that's usually the agent script terminating, and the fix belongs there rather than in the restart policy.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+        value = local.owner_token
+      }
+
+      # JSON containing the "wellknown" auth entry (keyed by opencode URL) so the
+      # startup script can write $HOME/.local/share/opencode/auth.json. This is
+      # what makes opencode fetch the remote config from the URL automatically.
+      env {
+        name  = "OPENCODE_OAUTH_AUTH_JSON"
+        value = local.opencode_auth
+      }
```
stephane-segning: 🔴 **Blocker (security) — a live user token in the pod spec.**

`OPENCODE_OAUTH_ACCESS_TOKEN` and `OPENCODE_OAUTH_AUTH_JSON` (which embeds the same token) are literal `env` values on `kubernetes_pod_v1`. That makes a valid Keycloak access token for a named human readable by:

- anyone with `get pod` in the `coder` namespace — `kubectl get pod -n coder <x> -o yaml`
- the workspace build's Terraform state, which coderd persists (state encryption is an enterprise feature)

The comment at L256-258 frames this as safe because no client secret or refresh token is present — but a bearer token the gateway accepts **is** the credential; that's the whole ADR-0021 boundary. And the careful `chmod 0600` on `auth.json` is undone by the env var it was read from.

**Fix:** deliver via `resource "coder_env"` — it reaches the agent over the agent API and never enters the pod spec or the k8s API. If it truly must be a pod env, use a per-workspace `kubernetes_secret_v1` + `value_from.secret_key_ref`.

While here: `CAMER_API_KEY` (L251) is dead — nothing reads it. Delete it along with `var.openai_api_key`.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+        limits = {
+          cpu    = "2"
+          memory = "4Gi"
+        }
+      }
+    }
+
+    dns_policy = "ClusterFirst"
+  }
+}
```
stephane-segning, line 507: **No PVC — the workspace is fully ephemeral.** There's no `kubernetes_persistent_volume_claim` and no `volume_mount` for `/home/coder`. Since `count = start_count`, every stop/start recreates the pod and wipes:

- `local.workdir` (`/home/coder/project`) — the user's actual work
- `~/.cache/opencode/node_modules/` — forcing a full plugin re-install
- `~/.cache/opencode-oauth2/` — the seeded token cache

Coder's canonical Kubernetes template always provisions a home PVC. Even for a task-oriented workspace this is surprising enough that the README should say so if it's deliberate.

Also missing on the pod spec:
- `automount_service_account_token = false` — otherwise the workspace carries the `coder` namespace default SA (see the L123 note on blast radius)
- a `security_context` (`run_as_non_root`, `allow_privilege_escalation = false`)
- the `com.coder.*` labels Coder's UI uses to associate pods with workspaces — only `app = coder-workspace` is set (L224-226), so every workspace looks identical in `kubectl` output

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,319 @@
...
+  workdir              = local.workdir
+  opencode_version     = "latest"
+  ai_prompt            = data.coder_task.me.prompt != "" ? data.coder_task.me.prompt : ""
+  report_tasks         = true
+  install_agentapi     = true
+  agentapi_version     = "v0.11.2"
+  web_app_display_name = "OpenCode"
+  icon                 = "/icon/opencode.svg"
+  config_json          = local.opencode_config
+  auth_json            = local.opencode_auth
```
stephane-segning: Two small ones:

- **L306** is a tautology — `x != "" ? x : ""` is just `x`. Simplify to `ai_prompt = data.coder_task.me.prompt`.
- **L313 duplicates the startup script.** The module writes `auth.json` from this arg *and* the startup script writes it at L174-176. Beyond the redundancy, `coder_script` resources from modules aren't ordered against `coder_agent.startup_script`, so which write lands last isn't guaranteed. Keep this one (it's the maintained path) and delete the startup-script block — which also disposes of the `$$HOME` bug for free.

terraform/coder/templates.tf
```diff
@@ -0,0 +1,35 @@
...
+  description     = "Simple Docker POC template (declarative delivery test, ADR-0120)"
+  organization_id = var.organization_id
+
+  versions = [{
+    name      = "v-${var.git_commit_sha}"
+    message   = "Initial declarative test version"
+    directory = "${path.root}/../../.coder/templates/poc-ubuntu"
+    active    = true
+  }]
+}
```
stephane-segning: 🔴 **Blocker — this file cannot `init`, `plan` or `apply`.**

`terraform/` doesn't exist on `main` (`git ls-tree origin/main -- terraform .coder` → empty) and this PR adds only `templates.tf`, so the directory has:

- no `terraform {}` / `required_providers` for `coderd`, and no `provider "coderd"` block (url + token)
- no `variable "organization_id"`, no `variable "git_commit_sha"` — both referenced here and below
- no `backend "s3"` — ADR-0120 specifies Hetzner Object Storage state (`ssegning-k8s-state`, key `coder/templates/terraform.tfstate`)
- **`directory = "${path.root}/../../.coder/templates/poc-ubuntu"` pointing at a directory that exists in neither this repo nor any open PR.** `plan` fails on this resource alone, before it ever reaches `opencode_task`.

There's also no Terraform job in `.github/workflows/`, which is why CI is green on this PR — nothing validates this file. Worth flagging explicitly since ADR-0120 makes the PR→plan→apply pipeline the delivery mechanism.

Cleanest split: drop this file from #928 and ship the template dir alone; land `terraform/coder/` (main.tf + variables.tf + backend + workflow) as its own PR on top of #913, adding `poc_ubuntu` only once `.coder/templates/poc-ubuntu` is actually committed.

stephane-segning (commented):
## Review — `opencode-task` template

The core idea is right: URL-driven config from `.well-known/opencode` instead of a hand-built provider block is exactly the ADR-0014/0042 contract, and seeding the **owner's own** token (rather than a shared service account) preserves ADR-0021 per-user `x-account-id` budget attribution. Nice.

Four things block merge, detailed inline. Summary:

### 1. `terraform/coder/templates.tf` cannot `init`, `plan` or `apply`
`terraform/` does not exist on `main` (`git ls-tree origin/main -- terraform .coder` → empty) and this PR adds only `templates.tf`. So there is no `required_providers`/`provider "coderd"`, no `var.organization_id`, no `var.git_commit_sha`, no `backend "s3"`, and **`coderd_template.poc_ubuntu` points at `.coder/templates/poc-ubuntu`, which exists in neither this repo nor any open PR** — `plan` fails on that resource alone. There's also no Terraform workflow in `.github/workflows/`, which is why CI is green here: nothing validates this file. A green pipeline proves nothing about this PR.

Also worth naming: this is stacked on #913 (ADR-0120), which is still open, and implements roughly a third of the repo layout that ADR specifies (`main.tf` + `variables.tf` + backend + PR→plan→apply workflow).

**Ask:** either drop `templates.tf` from this PR and ship the template dir alone, or land the full `terraform/coder/` scaffold and remove `poc_ubuntu` until that directory exists.

### 2. The `$$HOME` escaping bug is still live
It was fixed for the node block only — the `auth.json` block still has it. Verified locally against Terraform: `$$` is not an escape sequence (only `$${` is), so it renders literally and the shell expands `$$` to the PID. Details inline at `main.tf:174`.

### 3. The owner's live OIDC token is written into the pod spec
Readable via `kubectl get pod -n coder <x> -o yaml` and persisted in the workspace build's Terraform state. Inline at `main.tf:259`.

### 4. The TTL mitigation fails closed with no recovery
Seeding without `expiresAt` and without a refresh token means the workspace works for ~5 minutes, then every gateway call 401s and the plugin — having been told the token is permanently valid — never re-auths or prompts. The only recovery is a workspace rebuild. "Authenticate non-interactively" is the whole feature; a five-minute window isn't a working version of it, and shipping the trick without the Keycloak TTL change puts a template in the UI that looks like it works. Inline at `main.tf:192` with options (`coder_external_auth` is probably the idiomatic answer).

### Also worth fixing
- **No PVC** — the workspace is fully ephemeral; every stop/start wipes `/home/coder/project` and the plugin cache and re-runs the whole apt + nodesource + install cycle.
- **Four declared variables are never used** (`openai_base_url`, `keycloak_issuer_url`, `model`, `openai_api_key`) but the README documents them as functional knobs.
- **`.terraform.lock.hcl` is gitignored** and the provider constraints are unbounded `>=` floors — the template is not reproducible from git, which is the one thing ADR-0120 exists to guarantee.
- `auth.json` is written twice (module arg + startup script), unordered relative to each other.
- Leftover scratch comments and commented-out alternatives (`# This is what was missing!`, `# Option B (Alternative)`, the `config_path` fallback) — house rule is no dormant alternate paths.
- **Docs sweep not done**: `docs/integrations/coder-platform-integration.md` still documents the per-workspace client-credentials auth model this PR supersedes, and ADR-0120 is referenced nowhere but a `description` string.

### Suggested path
1. Merge #913, then split this: land the `terraform/coder/` scaffold as its own PR, drop `poc_ubuntu` until its directory exists.
2. Fix `$$HOME` (or delete the redundant block).
3. Move the token to `coder_env`; delete `openai_api_key`/`CAMER_API_KEY`.
4. Decide the TTL story — don't ship the `no expiresAt` trick as the answer.
5. Add the home PVC; delete or wire the unused variables; commit the lock file.

Happy to look again once the scaffold question is settled — that one decides whether this is one PR or two.

---
*Review by Claude Opus 5 at @stephane-segning's request. Every claim above was checked against the repo or reproduced locally; the `$$` behaviour was verified with an actual `terraform apply`.*

.coder/templates/opencode-task/.gitignore
```diff
@@ -0,0 +1,5 @@
+.terraform/
+.terraform.lock.hcl
```
stephane-segning: `.terraform.lock.hcl` should be **committed**, not ignored.

The lock file is what pins provider versions inside the template tarball the `coderd` provider content-hashes and ships to Coder. Ignoring it means every version build re-resolves providers, so the same git commit can produce different workspace behaviour — directly against ADR-0120's "reproducible from git" premise. See also the `>=` constraints at `main.tf:4-11`.

The other four entries are right.

.coder/templates/opencode-task/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Variables
+
+| Variable | Default | Description |
+|---|---|---|
+| `namespace` | `coder` | Kubernetes namespace for the workspace pod (override to `coder-flows` locally) |
+| `opencode_url` | `https://ai.camer.digital/opencode` | OpenCode server URL; remote config is fetched from `<url>/.well-known/opencode` |
+| `provider_key` | `camer-digital` | Provider/server id used as the oauth2 plugin cache filename (must match the provider key in the remote config) |
+| `model` | `adorsys-coder` | Default model name (models come from the remote config) |
+| `openai_base_url` | `https://api.ai.camer.digital/v1` | OpenAI-compatible API base URL |
```
stephane-segning: This table documents four variables that the template never reads — `opencode_url` and `provider_key` are genuinely wired, but `model`, `openai_base_url` (and `keycloak_issuer_url`, not listed) are declared and ignored. See `main.tf:53`.

While you're here: two things the README should state that it currently doesn't, because both will surprise the first user —

1. **The workspace has no persistent storage** — `/home/coder/project` is wiped on every stop/start (`main.tf:282`).
2. **The seeded token's lifetime is the Keycloak access-token TTL (~5 min by default)**, after which the gateway 401s with no in-workspace recovery (`main.tf:186`). If that TTL bump is a prerequisite, the README is where someone will look for it.

Worth also linking ADR-0120 from here — right now it's referenced nowhere but a `description` string in `templates.tf`.

benie-joy-possi (commented):
## Code Review — `feat(coder): per-workspace ServiceAccount auth for OpenCode (PR #928)`

Reviewed the full diff against `main` (`.coder/templates/opencode-task/main.tf`, README, ADR-0130, and the two integration docs). Overall the architecture is sound and well-documented — the per-workspace SA (`coder-<sub>.<plan>`), openresty sidecar injecting a projected SA token, and pure-CEL derivation via `kubernetesTokenReview` are a clean, non-forgeable, credential-light design. No blocking issues. Findings below, ordered by importance.

### P2 — Loopback sidecar is NOT isolated from the user's own container
`main.tf` binds the sidecar to `127.0.0.1:8080` and justifies it as "must not be reachable by other in-cluster workloads." True across pods, **but containers in the same pod share the network namespace**, so the user-controlled workspace container *can* reach the sidecar on localhost. This is by design (OpenCode uses it) and is **not a privilege escalation** — attribution is the user's own SA — and the ADR does acknowledge the shared netns with "credential secrecy (token read only in the sidecar) as the primary control, with a CiliumNetworkPolicy as defense-in-depth."
Please make this explicit and confirm the CiliumNetworkPolicy is actually shipped: the stronger phrasing in `coder-workspace-opencode-auth-flow.md` ("the user container cannot reach the gateway, so it cannot forge the header") is misleading given the shared netns. If isolation-from-the-owner were ever required, the proxy would need to be a *separate pod* (as the doc's own §3 correctly states). Recommend clarifying the security model in the template comments + flow doc.

### P2 — `docs/integrations/README.md` links a file that is not in this PR
The README change adds a row for `coder-workspace-opencode-web-link.md`, but that file is **untracked / not part of this PR**, while the two docs actually added here (`coder-workspace-opencode-auth.md`, `coder-workspace-opencode-auth-flow.md`) are **not** listed. After merge the README will contain a broken link and the new docs will be unreachable from the index. Fix: reference the two new docs (and move the web-link row to whichever PR adds that file).

### P2 — PR description vs. actual diff
Body states it "**Removes** `terraform/coder/templates.tf` (declarative-template scaffold)." That path does not exist on `main` nor in this diff, and this PR contains **zero deletions** (all 9 files are additions/modifications of existing READMEs). Update the description to avoid misleading reviewers about a removal that doesn't happen here.

### P3 — openresty Lua has no error handling on token read
```lua
local f = io.open("/var/run/secrets/gateway-token/token", "r")
local token = f:read("*a"); f:close()
```
`io.open` result isn't nil-checked; if the projected token file is ever absent/unreadable at request time this throws and every `/` request 500s with an opaque error. With projection + rotation the file should be present, but a graceful path (e.g., `return 503`) would be far more diagnosable. Minor, suggest.

### P3 — startup script runs a remote installer every boot
`curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash -` downloads/executes a remote script on each start, and node/npm are reinstalled every (ephemeral) boot. Acceptable given the deliberate no-PVC design, but a pinned base image carrying node + apt layers (or at least pinning the installer version) would reduce supply-chain surface and start time. Worth a follow-up note, not a blocker.

### P3 — Template dir is outside PR-time plan/validate coverage
The in-cluster `kubernetes` provider means `terraform validate` fails outside the cluster (documented in `main.tf`). Consider at least a CI step running `terraform fmt -check` and `tflint` on this dir so the "no automated tests" gap is partly closed. Also confirm the `~> 2.13` vs. lockfile `2.18.0` (constraint `>= 2.12.0, ~> 2.13`) is intentional.

### Non-blocking / noteworthy
- **Naming↔CEL coupling:** the whole scheme depends on the SA-name format `coder-<sub>.<plan>` and Authorino's CEL (in `ai-helm-values`) staying in lockstep. A silent mismatch would mis-attribute budgets. Cross-link the two and add a comment near the `sa_name` local pointing at the CEL.
- **`billing_plan` fallback `"free"`:** a missing mapper claim silently attributes the user as `free`. Reasonable default and documented (plan changes require rebuild), just flagging the silent-behavior aspect.
- **Workspace container has no `security_context`** (runs as root). Deliberate for an interactive workspace, but consider `runAsNonRoot`/capability drops where feasible.
- Great work on the ADR-0130, the honest "why not Caddy / plan-in-label / Rust-proxy" rationale, and the header-strip + loopback + read-only-FS hardening.

.coder/templates/opencode-task/main.tf
```diff
@@ -0,0 +1,377 @@
...
+        requests = {
+          cpu    = "500m"
+          memory = "1Gi"
+        }
+        limits = {
+          cpu    = "2"
+          memory = "4Gi"
+        }
+      }
+    }
```
stephane-segning, line 428: Make this configurable. Plus, for a simple opencode pod, that's too much.

stephane-segning (changes requested):

PR #937, opened 2026-08-06
feat(coder): LibreChat Coder agent — converse-delegated subagent + authenticated port-share (P1/P2 applied)

Guy-Ghis:
## 1. Summary

This PR implements the **LibreChat Coder agent (Ticket #832)** following the **converse→coder subagent-delegation** direction (ADR-0086/0088), and applies the maintainer's review feedback (P1 + P2s).

Changes:

- Registers the `coder` DB agent in `charts/librechat-app/values.yaml` `agentSeed.agents`, delegated by `converse` via `subagentNames` (idempotent `seed-agents.js`).
- Resolvable ids: `model: adorsys-coder-pro-internal` + `mcpServers: ["coder_mcp"]` (→ `sys__all__sys_mcp_coder_mcp`), both verified against the converse-provider model list + registered MCP servers.
- Port-sharing contract per ADR-0121.
- Restores `docs/adr/0086-...md` and `docs/integrations/coder-workspace-urls.md` to `origin/main` (the prior prompt-centric PR rewrote the ADR + deleted ~537 lines of hard-won Coder knowledge).

### Reviewer feedback applied (PR #937)
- **P1 (block):** HARD-set the `coder` agent to `share_level: authenticated` only. It must never create a `public` (unauthenticated) share — public exposure requires a human admin decision and is outside the agent's authority. Applied to the coder prompt (chart values), the live seed, and the workspace-urls contract.
- **P2a:** `charts/librechat-app/values.yaml` `agentSeed.agents` is the **canonical** fleet source; ADR-0086 + system-prompt doc reference it instead of re-declaring.
- **P2b:** documented the `coder_mcp` credential scope (shared deployment-level LibreChat OAuth client; per-user isolation via `X-USER-*` headers; keep non-admin/workspace-scoped; never an admin/org-wide Coder token).
- **P2c:** added `tools/check-agent-seed.sh` validating every `agentSeed.agents[*]` model id + `mcpServers[*]` → `sys__all__sys_mcp_<name>` token expansion.

It solves:

- https://github.com/ADORSYS-GIS/ai-helm/issues/832 (LibreChat Coder agent) — parent Epic #821

---

## 2. Intent

> Enable web-application scaffolding from LibreChat by giving the `coder` agent (delegated from `converse`) a secure, resolvable Coder integration that returns a live preview link.

---

### In Scope

- `coder` subagent under `converse`, `agentSeed.agents`, resolvable model/MCP ids.
- Port-share contract (authenticated-only).
[...]

PR #950, opened 2026-08-08
feat(coder): declarative delivery of the poc-ubuntu workspace template (ADR-0123)

### LINKED ISSUE #923: [Ticket]: Implement declarative Coder workspace templates (ADR-0120)
### Type

Feature

### Summary

We need to deliver our custom Coder workspace templates declaratively via the `coder/coderd` Terraform provider, per ADR-0120, because the current workflow relies on a manual `coder templates push` to the live server — a stateless, non-GitOps step that makes workspaces non-reproducible from git for a fresh cluster.

Build the actual workspace template **source** (`.coder/templates/{name}/`) and the **lifecycle** wiring (`terraform/coder/` `coderd_template` resources) so templates deploy automatically from git on merge to `main`, with no manual CLI push required.

Expected result:
> At least our first custom workspace template is reproducible from git: a fresh Coder cluster + Terraform remote state + credentials reproduces the template without any manual `coder templates push`.

### Intent

Make workspace templates a first-class, reviewable, version-controlled asset of the `ai-helm` monorepo. A developer edits template source in git, opens a PR (validated by `terraform plan`), merges, and the deploy workflow applies it to Coder. Drift is detectable via `terraform plan` and reversible via git — removing the manual, error-prone, non-auditable push path decided in ADR-0120.

### Source of truth (links)

- ADR: `docs/adr/0120-declarative-coder-template-management.md`
- Pattern (the *how*): `docs/patterns/coder-declarative-templates.md`
- Strategy spike (this ticket's predecessor): #834
- Parent epic: #822
- Provider: https://registry.terraform.io/providers/coder/coderd/latest

### Current Behavior

Coder workspace templates are deployed by an operator running `coder templates push <name> --directory …` against the live Coder server — imperative, no version control of the deployment, no rollback, no drift detection (documented in ADR-0120). The strategy is decided and documented, but no custom `.coder/templates/` source or `terraform/coder/` lifecycle exists yet.

### Expected Behavior

- Template **source** lives under `.coder/templates/{name}/` in the `ai-helm` monorepo.
- Template **lifecycle** lives in `terraform/coder/` (`coderd_template` resources with `versions[].directory` pointing at the source, `active = true`, Git-SHA-encoded version names, remote state on Hetzner Object Storage).
- A GitHub Actions workflow validates on PR (`terraform fmt` + `terraform validate` + `terraform plan`) and applies on merge to `main`.
- A fresh Coder cluster + Terraform state + credentials reproduces templates with **no manual `coder templates push`**.

### Acceptance Criteria

- [ ] Given a new Coder cluster, when `terraform apply` is run against `terraform/coder/`, then the template is created autonomously with a managed version (no manual push).
- [ ] Given a change to template source, when merged to `main`, then the deploy workflow creates a new Coder version (content-hash detected) with a new SHA-encoded name and marks it active.
- [ ] Given an existing/pushed template, when adopted, then it can be imported via `terraform import` and reach "no drift" on `terraform plan`.
- [ ] Error cases are handled safely: a `terraform plan` failure or a missing required version (`active = true`) fails the PR/build rather than silently deploying.
- [ ] Existing behavior is not broken: no regression to the Coder deployment or other charts.
- [ ] Relevant tests/validations are added (Terraform `fmt`/`validate` in CI; template source contains `main.tf`/`variables.tf`).
- [ ] Verification evidence is provided (see Test Plan / Verification).

### Out of Scope

- Coder **server-level** administration/config (tracked in the runbook ticket; e.g. provisioners, licensing, global settings).
- Re-evaluating the delivery strategy (already decided in ADR-0120).
- Delivering *all* templates in one pass — start with **one** template as the pilot proof, then extend.

### Technical Context

- Provider `coder/coderd` v0.0.22 (requires Coder ≥ 2.10.1; we run v2.34.6). The `coderd_template` resource supports `versions[].directory`, `active`, `tf_vars`, `provisioner_tags`, `acl` (Enterprise), TTL/dormancy, deprecation.
- **Version detection:** the provider content-hashes the directory (`sha256`) at plan time and creates a new version when it changes; version **names** must change when contents/`tf_vars` change (encode the Git SHA — see the pattern doc).
- **State backend:** `backend "s3"` against Hetzner Object Storage — bucket `ssegning-k8s-state`, key `coder/templates/terraform.tfstate`, with Ceph-RGW flags (`endpoint nbg1.your-objectstorage.com`, `skip_*`, `force_path_style`). Credentials from `ssegning-aws` (`prod/meta/test-app`) via `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`.
- Coder credentials: API token with template-admin scope; `CODER_URL`/`CODER_TOKEN` as GitHub Actions secrets.
- Relevant files: `terraform/coder/{main.tf,templates.tf,variables.tf}`, `.coder/templates/{name}/`, `.github/workflows/validate-coder-templates.yaml` + `deploy-coder-templates.yaml`.
- Known gotcha: some Enterprise attributes (e.g. `max_port_share_level`) can need a second apply after first creation (two-phase apply).

### Risks

| Risk | Impact | Mitigation |
|------|--------|-----------|
| Terraform state corruption / lock contention | High | Remote S3 state + versioning on `ssegning-k8s-state`; pin provider version |
| Accidental template deletion | High | PR review gating; `prevent_destroy`/`protect` where appropriate; validate before apply |
| Provider is young (v0.0.x) breaking change | Medium | Pin `required_providers` version; test upgrades in the validate workflow |
| Coder API rate limiting during applies | Low/Med | Terraform built-in retry; low-frequency triggers |
| Values/creds not provisioned first | High (blocker) | Set `CODER_URL`/`CODER_TOKEN` secrets and confirm S3 creds before first apply |

### Test Plan

- **Validate (PR):** `terraform fmt -check -recursive terraform/coder`; `terraform init -backend=false && terraform validate`; assert each `.coder/templates/*/` has `main.tf` (+ `variables.tf` where applicable).
- **Plan:** `terraform plan -var="git_commit_sha=<sha>"` must succeed and show intended changes on a review build.
- **Manual/e2e (pilot):**
  ```bash
  cd terraform/coder
  terraform init
  terraform plan -var="git_commit_sha=$(git rev-parse HEAD)" -out=tfplan
  terraform apply tfplan
  coder templates versions list <name>   # expect new SHA version, active
  coder create test-ws --template <name> # expect workspace provisions
  coder delete test-ws
  ```
- **Regression:** confirm no other charts/ArgoCD Apps affected (docs/ADR only + new `terraform/coder/`, `.coder/`, workflow files).
- **Expected result:** all validate job steps green; apply creates the template; plan returns to "no changes" on a second run (idempotent).

### Verification evidence

- Test output: paste `terraform validate` / `terraform plan` output; CI workflow run links.
- Manual verification steps: record the `coder templates versions list <name>` and workspace create/ssh/delete results.
- Logs / metrics checked: deployment workflow logs; Coder API template list shows the template with the expected active version.
- Known remaining limitations: (to be filled by the accountable owner) — e.g. two-phase apply for a given Enterprise attribute if hit, or multi-template rollout deferred.

### Human accountable owner

@benie-joy-possi

### AI Usage Declaration

- Drafting the ticket
- Writing documentation

### Human verification completed

- [x] I understood the intent
- [x] I checked the source of truth
- [ ] I reviewed all AI-generated text/code
- [ ] I verified the implementation manually
- [ ] I verified the tests
- [ ] I checked for hallucinated assumptions
- [x] I documented remaining risks
- [x] I am the accountable owner and accept responsibility for this ticket.

benie-joy-possi:
Closes #923

## 1. Summary

This PR changes:

- Adds `.coder/templates/poc-ubuntu/` — a minimal Kubernetes pod workspace (`codercom/enterprise-base:ubuntu` in the `coder` namespace) delivered declaratively via the `coder/coderd` Terraform provider, instead of a manual `coder templates push`.
- Adds `terraform/coder/` — the declarative **lifecycle** (`coderd_template` with `versions[].directory` pointing at the template source, `message`/`active`, Git-SHA-encoded version name), and an S3 state backend on Hetzner Object Storage.
- Adds GitHub Actions workflows: **validate** (PR-gate: `terraform fmt` + `validate`) and **deploy** (applies on merge to `main`).
- Indexes the new pattern/ADR wiring; docs updated for import ID and fresh-cluster empty-state.

It solves:

- #923 (parent epic #822) — implements ADR-0123.

---

## 2. Intent

The intent of this PR is:

> Make the first custom Coder workspace template (poc-ubuntu) reproducible from git: a fresh Coder cluster + Terraform remote state + credentials reproduces the template with no manual `coder templates push`, per ADR-0123.

---

### In Scope

- `.coder/templates/poc-ubuntu/` (main.tf, README.md, .gitignore) — Kubernetes pod workspace
- `terraform/coder/` (main.tf, variables.tf, templates.tf, README.md, .gitignore) — `coderd_template` lifecycle + S3 state backend
- `.github/workflows/validate-coder-templates.yml` (PR-gate) + `deploy-coder-templates.yml` (apply on merge)
- ADR-0123 review corrections applied (S3 flag names, `use_lockfile`, `message` in `versions`, import-ID docs)

### Out of Scope

- Server-level Coder admin/config (provisioners, licensing) — runbook-tracked
- Delivering every template — this is the pilot (poc-ubuntu) proof; other templates follow
- The `opencode-task` template (delivered separately in PR #928)

---

## 4. Verification

I verified this change by:

- [x] Running automated tests
- [ ] Running manual tests
- [ ] Checking logs
- [ ] Checking metrics
- [x] Testing error cases

Commands run:

```bash
terraform fmt -check terraform/coder
cd terraform/coder && terraform init -backend=false && terraform validate
cd .coder/templates/poc-ubuntu && terraform fmt -check && terraform init -backend=false && terraform validate
```

Results:

```text
terraform/coder:           Success! The configuration is valid.
.coder/templates/poc-ubuntu: valid config (note: in-pod Kubernetes service-account
  file() paths only resolve inside a Coder provisioner pod — same as PR #928).
terraform fmt:             pass (canonical formatting applied).
```

---

## 5. Screenshots / Evidence

image

---

## 6. Risk Assessment

Risk level:

* [ ] Low
* [x] Medium
* [ ] High

Potential risks:

* Coder admin token + S3 credentials must exist as GitHub Actions secrets (ESO → cluster doesn't reach Actions)
* `runs-on: adorsys-gis-runner` — confirms the ARC self-hosted runner label is registered for this repo (GitHub-hosted runners are billing-blocked org-wide, CLAUDE.md)
* Provider is young (`v0.0.x`); pinned `~> 0.0` and validated in CI

Mitigation:

* Secrets documented; validate workflow catches parity/`fmt` drift before merge
* `prevent`-style gating via PR review + `terraform plan` before apply

---

## 7. AI Usage Declaration

AI was used for:

* [x] Generating code
* [ ] Refactoring
* [ ] Generating tests
* [x] Drafting documentation
* [ ] Reviewing the diff
* [ ] Not used

Human verification:

* [x] I checked generated tests manually

---

## 8. Reviewer Focus

Please focus your review on:

* [x] Correctness
* [x] Architecture
* [x] Security
* [ ] Performance
* [x] Tests
* [x] Maintainability
* [x] Product intent
* [x] Edge cases


.coder/templates/poc-ubuntu/main.tf
```diff
@@ -0,0 +1,79 @@
...
+    sudo apt-get update
+    sudo apt-get install -y git curl
+  EOT
+}
+
+resource "kubernetes_pod_v1" "workspace" {
+  count = data.coder_workspace.me.start_count
+
+  metadata {
+    name      = "coder-${data.coder_workspace.me.id}"
```
stephane-segning: Can we use "coder-workspace--*" instead ?

stephane-segning (changes requested):

terraform/coder/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  type        = string
+  description = "Coder API token with template-admin scope"
+  sensitive   = true
+}
+
+variable "organization_id" {
+  type        = string
+  description = "Coder organization ID"
+  default     = "default"
+}
```
stephane-segning: 🔴 **`"default"` is not a valid `organization_id` — this fails at apply.**

The provider types this field as a UUID, not a free string:

```go
// internal/provider/template_resource.go
OrganizationID  UUID  `tfsdk:"organization_id"`      // :65
orgID := data.OrganizationID.ValueUUID()             // :580
```

**Fix: delete the variable and the attribute.** The provider already handles the default org itself:

```go
if data.OrganizationID.IsUnknown() {                 // :566
    data.OrganizationID = UUIDValue(r.data.DefaultOrganizationID)
}
```

So omitting `organization_id` from `coderd_template` is the supported way to say "the default org" — and it's more correct than hardcoding any UUID, since it follows whatever the token's deployment actually uses.

Why this one is worth catching pre-merge: **no gate in this PR sees it.** `terraform validate` accepts any string, the PR workflow never runs `plan`, and the first `apply` is on `main`. It's the same shape as the `description`/`message` bug in #913 — a plausible-looking string in a typed field, invisible to validation.

One thing that makes the mistake *easy*: `README.md:59` uses `default` correctly as the org **name** for the import ID. Same word, different identifier, one screen apart. Worth one sentence there distinguishing them.

terraform/coder/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  type        = string
+  description = "Coder deployment URL"
+  sensitive   = true
+}
+
+variable "coder_token" {
+  type        = string
+  description = "Coder API token with template-admin scope"
+  sensitive   = true
+}
```
stephane-segning, line 10: `coder_url` as `sensitive = true` is over-marking — it's a deployment URL, not a credential, and marking it sensitive redacts it from *all* plan/apply output, so a failing deploy won't even show you which Coder it was talking to. Drop `sensitive` here; keep it on `coder_token` (correct).

If the URL is considered non-public, the `sensitive` flag isn't what protects it anyway — the value still lands in state.

stephane-segning (commented):
## Review — declarative delivery of `poc-ubuntu`

This is the scaffold that #928 assumed and neither it nor #913 provided, and it's a clear step up. Several things I flagged on #928 are fixed here on their own merits:

- **`provider "kubernetes" {}` (empty) is correct** — I checked the provider docs, and it detects in-cluster via `KUBERNETES_SERVICE_HOST`/`KUBERNETES_SERVICE_PORT`, so no attributes are needed inside the provisioner pod and it still falls back to `KUBECONFIG` locally. The code comment is accurate. This is strictly better than #928's `file("/var/run/secrets/...")`, which can't be validated outside the cluster.
- `message` (not `description`) inside `versions`, with an inline warning ✅
- Every S3 backend correction landed: `endpoints.s3`, `use_path_style`, `use_lockfile`, `skip_s3_checksum`, with both "verify against Ceph-RGW" caveats preserved ✅
- `terraform plan -out=tfplan` → `apply tfplan` (applies exactly what was planned) ✅
- `data.coder_provisioner.me.arch/os` instead of hardcoded `amd64` ✅

Two blockers, both of which land *after* merge — which is the problem.

### 🔴 1. `organization_id = "default"` is not a valid value
`variables.tf:16` defaults it to the literal string `"default"`. In the provider, that field is a **UUID**:

```go
OrganizationID  UUID  `tfsdk:"organization_id"`   // template_resource.go:65
orgID := data.OrganizationID.ValueUUID()          // :580
```

`"default"` will not parse. The correct move is to **omit the attribute entirely** — the provider already falls back to the deployment's default org (`:566-567`, `UUIDValue(r.data.DefaultOrganizationID)`).

What makes this sharp rather than a typo: **nothing in this PR catches it.** `terraform validate` sees a valid string, the PR gate never runs `plan`, and there's no `apply` until merge — so this surfaces as a red `deploy` job on `main`. Same failure class as the `description`/`message` bug in #913: a plausible string in a typed field.

(Note this is a *different* thing from the import ID in `terraform/coder/README.md:59`, where `default` really is the org **name**. Both can be right; they're different identifiers. Worth a sentence in the README saying so, since having `default` mean two things one screen apart invites exactly this mistake.)

### 🔴 2. ADR-0123 is already taken — again
`main` currently has:

```
0122-self-hosted-code-interpreter.md      Accepted 2026-08-07
0123-mlops-dedicated-cnpg-cluster.md      Accepted 2026-08-09
```

#913 has now been renumbered into a collision **twice** (0120 → 0122 → 0123, each taken by the time it landed). Next free is **0124**. This PR hardcodes `ADR-0123` in seven places — both workflow headers, both READMEs, `templates.tf`, and the template README's file link — so each renumber is a multi-file sweep across two PRs.

The structural fix is to stop hardcoding it: reference the ADR **once** (in `terraform/coder/README.md`) and drop the number from workflow comments and resource descriptions, which don't need it. The root cause is that a long-lived branch can't reserve a number in a repo merging ADRs weekly — worth landing #913 first and rebasing this on it.

### 🟠 3. `runs-on: adorsys-gis-runner` is an unverified deviation
Both workflows use it, citing CLAUDE.md's org-wide billing block. But:

- **`ai-helm` is a public repo** — GitHub-hosted runners are free for public repos and unaffected by that block.
- **All six existing workflows in this repo use `ubuntu-latest`** and pass (commit-lint, helm-lint, security, dashboards-drift, release-please, publish-charts-oci).
- `GET /orgs/ADORSYS-GIS/actions/runners` returns `total_count: 0`. That's consistent with ARC's JIT registration, so it isn't proof the label is missing — but it isn't proof it works either.

The failure mode is the bad one: an unmatched label doesn't fail, it **queues forever**. On the `deploy` workflow that means the template silently never ships and nothing goes red — and CLAUDE.md already records a 4-day runaway from exactly this family of ARC problem. Use `ubuntu-latest` to match the rest of the repo, or prove the label with a throwaway job first.

### 🟠 4. The PR gate doesn't gate the thing this PR adds
`validate-coder-templates.yml` runs `fmt`/`init`/`validate` against `terraform/coder` only. For `.coder/templates/*` it checks that `main.tf` **exists**. So a syntactically broken template passes the gate and fails at deploy — or worse, creates a broken version in Coder. Details inline at `validate:46`.

Also worth noting: ADR-0123 specifies the PR gate as "validates via `terraform plan`". A plan needs credentials and state, so what's implemented is `validate` only. That's a reasonable call — but it's a deviation from the ADR this PR implements, and it's the reason blocker #1 reaches `main`.

### 🟠 5. No concurrency group on `deploy`
Two merges in quick succession start two concurrent `terraform apply`s. `use_lockfile` protects the *state*, but the second run fails on lock contention. Needs a concurrency group with **`cancel-in-progress: false`** — cancelling mid-apply is how you get an orphaned S3 lock that blocks every later run.

### 🟠 6. The reproducibility pilot isn't reproducible
Both `.gitignore`s exclude `.terraform.lock.hcl`, and `poc-ubuntu/main.tf:3-5` declares the `coder` provider with **no version constraint at all**. For a PR whose stated purpose is "reproducible from git," the pilot resolves providers fresh on every version build. Commit both lock files and constrain the `coder` provider.

### Smaller items
- **`terraform_version: 1.11` is unquoted** and parses as a YAML float. It survives *by luck* — `1.10` would become `1.1`. Quote it. Same scalar-coercion family the repo already documents.
- `hashicorp/setup-terraform@v3` → **v4.0.1** is current.
- `fetch-depth: 0` + `git rev-parse HEAD` → `${{ github.sha }}` is already there; drop the full clone.
- `coder_url` marked `sensitive` — it's a URL, and this redacts it from all plan output.
- `git_commit_sha` defaults to `"dev"` — a silent version-name collision if the var is ever not passed.
- Template: `restart_policy = "Always"` (Coder's own K8s template uses `Never`), no `automount_service_account_token = false`, and `apt-get install git curl` installs what `enterprise-base:ubuntu` already ships.
- `poc-ubuntu/README.md:13` — backtick inside the link text: `` [`terraform/coder](...)` ``.

### Verdict
The architecture is right and the ADR corrections all landed faithfully. Before merge: drop `organization_id` (or pass a real UUID), settle the ADR number, and decide the runner question — those three are the difference between this deploying and this failing silently on `main`.

---
*Review by Claude Opus 5 at @stephane-segning's request. The `organization_id` type was read from the provider's `template_resource.go`, the in-cluster provider behaviour from the Kubernetes provider's own docs, and the runner/visibility facts from the GitHub API — not inferred.*

terraform/coder/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+  type        = string
+  description = "Coder organization ID"
+  default     = "default"
+}
+
+variable "git_commit_sha" {
+  type        = string
+  description = "Git commit SHA used in the template version name"
+  default     = "dev"
+}
```
stephane-segning, line 15: `default = "dev"` is a quiet footgun. Version names must be unique per content change (that's the whole reason the SHA is encoded), so if the var is ever not passed — a local `apply`, a workflow edit that drops the `-var` flag — every build becomes `v-dev` and collides with the previous one.

Since the only supported callers always pass it, I'd make it **required** (no default) so the omission fails loudly at plan instead of silently producing a colliding name. If you want local ergonomics, `v-dev-${timestamp()}` or reading it from the environment is safer than a fixed string.

.github/workflows/deploy-coder-templates.yml
```diff
@@ -0,0 +1,68 @@
...
+  # Map Actions secrets to Terraform vars via TF_VAR_ (coder_url/coder_token are
+  # required vars with no default — plain env names would NOT be read by Terraform).
+  TF_VAR_coder_url: ${{ secrets.CODER_URL }}
+  TF_VAR_coder_token: ${{ secrets.CODER_TOKEN }}
+  AWS_ACCESS_KEY_ID: ${{ secrets.SSEGNING_AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.SSEGNING_AWS_SECRET_ACCESS_KEY }}
+
+jobs:
+  deploy:
+    runs-on: adorsys-gis-runner   # GitHub-hosted runners are billing-blocked org-wide (CLAUDE.md)
```
stephane-segning: 🟠 **Two things on this job header.**

**1. `adorsys-gis-runner` looks like an unverified deviation.** The comment cites CLAUDE.md's org-wide billing block, but:
- **`ai-helm` is public** (`gh api repos/ADORSYS-GIS/ai-helm` → `visibility: public`), and GitHub-hosted runners are free for public repos regardless of org billing.
- **All six existing workflows here use `ubuntu-latest`** and pass.
- `GET /orgs/ADORSYS-GIS/actions/runners` → `total_count: 0`. Not conclusive (ARC registers JIT), but not reassuring either.

The failure mode is the dangerous one: an unmatched `runs-on` label doesn't error, it **queues indefinitely**. On the apply path that means the template silently never deploys and no check ever goes red. CLAUDE.md already records a 4-day ARC runaway from this family. Either match the repo (`ubuntu-latest`) or prove the label with a throwaway job before this workflow depends on it. Same applies to `validate-coder-templates.yml:23`.

**2. No `concurrency` group.** Two merges close together start two concurrent `terraform apply`s. `use_lockfile` protects state integrity, but the second run fails on lock contention rather than waiting its turn.

```yaml
concurrency:
  group: deploy-coder-templates
  cancel-in-progress: false   # never cancel a running apply
```

`cancel-in-progress: false` is load-bearing here — cancelling mid-`apply` is precisely how you end up with an orphaned S3 lock that blocks every subsequent run until someone clears it by hand. (The `true` on the validate workflow is correct — cancelling a `validate` is free.)

.github/workflows/deploy-coder-templates.yml
```diff
@@ -0,0 +1,68 @@
...
+          fetch-depth: 0
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: 1.11   # >= 1.10 for backend use_lockfile (S3-native locking)
+
+      - name: Set Git SHA
+        id: sha
+        run: echo "sha=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
```
stephane-segning: `fetch-depth: 0` clones the entire history so that `git rev-parse HEAD` can read a SHA that Actions already hands you as `${{ github.sha }}`. On a repo this size that's a real chunk of the job's wall time for nothing.

```yaml
- name: Checkout
  uses: actions/checkout@v7

# ...then use ${{ github.sha }} directly in the plan step
```

That removes the `fetch-depth`, the `Set Git SHA` step, and the `steps.sha.outputs.sha` indirection. (For a `push` event `github.sha` is the merge commit on `main`, which is what you want in the version name.)

.github/workflows/deploy-coder-templates.yml
```diff
@@ -0,0 +1,68 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v3
+        with:
+          terraform_version: 1.11   # >= 1.10 for backend use_lockfile (S3-native locking)
```
stephane-segning: **`terraform_version: 1.11` is unquoted, so YAML parses it as a float.** It happens to survive round-tripping here — but `1.10` would become `1.1`, which is a different (and unavailable) version. Given this repo already has a documented history with YAML scalar coercion (the ADR-0109 `"n": 1` boolean trap, and Helm turning a large int into `3.3554432e+07`), this is worth fixing on sight rather than when it bites:

```yaml
terraform_version: "1.11"
```

Also `hashicorp/setup-terraform@v3` is a major behind — **v4.0.1** is current. Same in `validate-coder-templates.yml:29`. Not urgent, but the repo generally tracks current majors (`actions/checkout@v7`, `azure/setup-helm@v5`, `actions/cache@v6`).

One more worth considering: this job holds `CODER_TOKEN` plus write-capable S3 credentials, and repo secrets are readable by any workflow on `main`. A GitHub `environment:` with required reviewers would scope them to this deploy. Optional, but cheap.

.github/workflows/validate-coder-templates.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          cd terraform/coder
+          terraform validate
+
+      - name: Check template source files
+        run: |
+          for d in .coder/templates/*/; do
+            [ -f "$d/main.tf" ] || { echo "ERROR: missing main.tf in $d"; exit 1; }
+            [ -f "$d/variables.tf" ] && echo "ok: $d" || echo "warn: no variables.tf in $d"
+          done
```
stephane-segning: 🟠 **The PR gate doesn't actually validate the templates.** This step only checks that `main.tf` *exists*. So a template with a syntax error, an unknown resource, or a bad reference sails through the PR gate and fails at deploy time on `main` — or, worse, `coderd` accepts the directory and you get a broken template version in Coder.

That matters more than usual here because the template directory is the thing this PR exists to deliver, and `terraform/coder` (which *is* validated) contains no template logic at all.

```yaml
- name: Validate template sources
  run: |
    terraform fmt -check -recursive .coder/templates
    for d in .coder/templates/*/; do
      terraform -chdir="$d" init -backend=false
      terraform -chdir="$d" validate
    done
```

The PR body says you ran exactly this locally — it should be the gate, not a habit.

Two smaller things in the current step:
- `fmt -check` at L34 covers `terraform/coder` only, so template formatting drifts unchecked.
- The `variables.tf` check prints `warn: no variables.tf` for `poc-ubuntu`, since it declares `variable "namespace"` inline in `main.tf` (which is fine and idiomatic). A check that warns on the repo's only template is noise — either drop it or make it a real convention and follow it.

.coder/templates/poc-ubuntu/main.tf
```diff
@@ -0,0 +1,79 @@
...
+  required_providers {
+    coder = {
+      source = "coder/coder"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.23"
+    }
+  }
+}
```
stephane-segning, line 12: **The `coder` provider has no version constraint at all**, and `.gitignore` excludes `.terraform.lock.hcl` (both here and in `terraform/coder/`). So the pilot whose entire purpose is proving "reproducible from git" resolves providers fresh on every version build — the same commit can produce different workspace behaviour a month apart.

The `coderd` provider is a `v0.0.x` that the ADR itself flags as young, which makes the lock file more valuable here, not less:

```hcl
coder = {
  source  = "coder/coder"
  version = "~> 2.13"
}
```

plus dropping `.terraform.lock.hcl` from both `.gitignore`s. The other four entries in each are right — it's only the lock line that works against the goal.

.coder/templates/poc-ubuntu/main.tf
```diff
@@ -0,0 +1,79 @@
...
+data "coder_workspace" "me" {}
+data "coder_workspace_owner" "me" {}
+
+provider "coder" {}
+
+# Coder's provisioner job runs INSIDE the cluster, so the hashicorp/kubernetes
+# provider auto-detects the in-cluster service-account config — no explicit
+# host/token/ca needed here. For local `validate`/`plan` it falls back to
+# KUBECONFIG or ~/.kube/config, so it doesn't fail outside the pod.
+provider "kubernetes" {}
```
stephane-segning, line 24: ✅ This is right, and worth saying so explicitly since it's a direct improvement on #928's approach.

I checked the provider docs to be sure the comment isn't wishful:

> The provider uses the `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` environment variables to detect when it is running inside a cluster, so in this case you do not need to specify any attributes in the provider block

So the in-cluster claim holds, *and* the local-fallback claim holds — which is why this template can be `validate`d in CI while #928's `file("/var/run/secrets/kubernetes.io/serviceaccount/token")` cannot. If #928 lands, it should adopt this block verbatim.

.coder/templates/poc-ubuntu/main.tf
```diff
@@ -0,0 +1,79 @@
...
+
+    container {
+      name    = "workspace"
+      image   = "codercom/enterprise-base:ubuntu"
+      command = ["/bin/sh", "-c", coder_agent.main.init_script]
+
+      env {
+        name  = "CODER_AGENT_TOKEN"
+        value = coder_agent.main.token
+      }
```
stephane-segning, line 69: Three pod-spec notes. All are softer than they'd be on a production template — the README is clear this is a mechanism proof — but two are cheap enough to just fix:

- **`restart_policy = "Always"`** — Coder's own Kubernetes template uses `Never`, so a workspace whose startup script fails surfaces as *stopped* rather than as a pod that restart-loops and re-runs `apt-get update` every cycle. Since the whole point is observing whether delivery worked, a loud failure is more useful here than a self-healing one.
- **`automount_service_account_token = false`** is missing, so the workspace carries the `coder` namespace default ServiceAccount. One line, no downside for a POC.
- **No PVC**, so the pod is fully ephemeral. Fine for a proof — but the README should say so, since "minimal template" and "loses your files on restart" aren't the same claim, and this template is the one people will copy.

Also minor: `apt-get install -y git curl` at L38-39 installs two things `codercom/enterprise-base:ubuntu` already ships. Harmless, but it makes every workspace start wait on an `apt-get update` for no gain.

terraform/coder/README.md
```diff
@@ -0,0 +1,73 @@
+# Declarative Coder templates — test wiring
+
+This directory is the **declarative lifecycle** for Coder templates (ADR-0123).
+The template **source** lives in `.coder/templates/`; this Terraform manages the
+template/version in Coder via the `coder/coderd` provider.
```
stephane-segning, line 5: 🔴 **ADR-0123 is taken on `main`** — by *"Give the mlops namespace its own dedicated CNPG cluster"* (Accepted 2026-08-09). `0122` is also gone (*Self-Hosted LibreChat Code Interpreter*, 2026-08-07). Next free is **0124**.

#913 has now been renumbered into a collision twice — `0120` → `0122` → `0123` — each number taken by the time the branch caught up. That's not a mistake anyone is making; it's structural, because a long-lived branch can't reserve a number in a repo merging ADRs weekly.

This PR hardcodes `ADR-0123` in **seven** places:
- `terraform/coder/README.md:3` and `:22` and `:70`
- `.coder/templates/poc-ubuntu/README.md:4` (including the **filename** in the link target)
- `terraform/coder/templates.tf:4` (resource `description`, which ships to Coder's UI)
- both workflow header comments

Suggestion beyond just renumbering: **cite the ADR once**, here, and drop the number from the workflow comments and the `coderd_template` description — neither needs it, and the resource description is user-visible in Coder, where an internal ADR number means nothing. That turns the next renumber into a one-line change instead of a two-PR sweep.

Cleanest sequencing is still: land #913 (as 0124), then rebase this on it so the link target actually resolves.

.coder/templates/poc-ubuntu/README.md
```diff
@@ -0,0 +1,13 @@
...
+decided in [ADR-0123](../../../docs/adr/0123-declarative-coder-template-management.md).
+
+It provisions a single Kubernetes pod based on
+`codercom/enterprise-base:ubuntu` (in the `coder` namespace) and runs a
+`coder_agent` with a small startup script (installs `git`, `curl`). It is
+intentionally **not** a production template — it exists to prove that a template
+can be delivered from git via the `coder/coderd` Terraform provider instead of
+`coder templates push`.
+
+The template lifecycle is managed from [`terraform/coder](../../../terraform/coder/README.md)`.
```
stephane-segning: Malformed link — the backtick is inside the link text and the closing backtick landed after the URL:

```markdown
The template lifecycle is managed from [`terraform/coder](../../../terraform/coder/README.md)`.
```

Should be:

```markdown
The template lifecycle is managed from [`terraform/coder`](../../../terraform/coder/README.md).
```

As written GitHub renders a link labelled `` `terraform/coder `` followed by a stray backtick. Also note the ADR link at L4 embeds the ADR **filename**, so it breaks on renumber — see the `terraform/coder/README.md` thread.

stephane-segning (commented):
## Re-review — all 12 findings fixed, and the gate now proves itself

I re-checked every item against the diff rather than the summary table. All of them landed, and several went further than asked:

| Finding | Verified |
|---|---|
| `organization_id = "default"` (UUID field) | Variable **and** attribute removed; `templates.tf:3-4` explains why in-line |
| ADR number | **0124** everywhere — and #913 merged as `0124-declarative-coder-template-management.md` at 07:46Z today, so the link resolves |
| "cite the ADR once" | Adopted — number dropped from both workflow bodies and the user-visible `coderd_template` description |
| `adorsys-gis-runner` | → `ubuntu-latest`, with the public-repo reason in the comment |
| PR gate didn't validate templates | Per-dir `init -backend=false` + `validate`, plus `fmt -check` on `.coder/templates` |
| No concurrency on deploy | Added, with `cancel-in-progress: false` and the orphaned-lock rationale |
| Not reproducible | Both lock files committed, `coder ~> 2.13`, `coderd ~> 0.0.3`, and a `.gitignore` comment explaining the deliberate non-ignore |
| YAML float | `"1.15"` quoted — and 1.15 is real (v1.15.8) |
| `setup-terraform@v3` | → `v4.0.1` |
| `fetch-depth: 0` + SHA step | Both gone, `${{ github.sha }}` used directly |
| `sensitive` on `coder_url` | Removed; kept on `coder_token` |
| `git_commit_sha` default | Now required |
| Pod spec | `Never`, `automount_service_account_token = false`, redundant `apt-get` dropped, name → `coder-workspace-*` |
| Markdown/docs | Link fixed, ephemeral-PVC warning added, and the "`default` means two things" note is clearer than what I suggested |

**The important part: the gate is self-testing.** I pulled the `validate` job log rather than trusting the green check — it really installed `coderd v0.0.22`, `coder v2.18.0`, `hashicorp/kubernetes v3.2.1` from the committed locks and validated both directories:

```
validating .coder/templates/poc-ubuntu/
- Installing coder/coder v2.18.0...
- Installing hashicorp/kubernetes v3.2.1...
Success! The configuration is valid.
```

That also settles a question I'd have raised otherwise — the locks resolve cleanly on `linux_amd64` in CI, so no cross-platform `terraform providers lock` work is needed.

**No blockers.** Three small things below, one of which I'd fix before merge.

---

## Meta-review of the other review

Since this feedback loops back into the assistant, the calibration notes matter as much as the findings.

**Sound:** the shared-state-bucket / least-privilege point (P3) is a good catch worth carrying forward, and "no `terraform plan` in the PR gate" (P3) is real — I raised it too.

**P1 (auto-apply on merge) — right topic, mis-anchored and over-weighted.** "Push to main equals deploy" is not a risk this PR introduces; it is the platform's **documented, deliberate** posture. ADR-0055 says so in as many words — the whole ArgoCD fleet auto-syncs from `main`, and immutability was abandoned on purpose. Calling it "the most important item here" targets the implementation of ADR-0124 rather than the decision, and the fix it proposes (plan-only + manual approval) would contradict both ADRs. There *is* something genuinely new to say — a template-admin token in Actions secrets, executing in-cluster, is a new credential inside that existing trust boundary — but that's an ADR-amendment conversation, not a P1 on this diff.

**P2 (`CODER_AGENT_TOKEN` in pod env) — right fact, wrong mechanism.** Credit where due: I checked #937 and `share_level: "public"` really is in there, so the cross-PR reference is accurate, not invented. But the inference doesn't hold. `share_level` exposes a **`coder_app`'s port** on a subdomain; it does not expose the pod's environment. And `poc-ubuntu` declares **no `coder_app` at all**, so nothing in this template is shareable by any level. The finding also concedes it's "the standard Coder pattern" without offering an alternative — because there isn't one: the agent token must reach the agent for the workspace to function.

**P1's `actions/checkout@v7` remark and P3 (tags not SHAs) contradict this repo's convention.** All six pre-existing workflows use tag refs for marketplace actions (`checkout@v7`, `setup-helm@v5`, `cache@v6`, `trivy-action@v0.36.0`); only the cross-org *reusable workflows* are SHA-pinned. Singling out this one workflow isn't calibrated to the repo — if SHA-pinning is wanted it's a repo-wide policy change, not a note on this PR.

**P3 (relaxed S3 flags) is partly thin.** `skip_credentials_validation` and `skip_metadata_api_check` are *required* for any non-AWS S3 endpoint — they aren't a relaxation with a tradeoff, they're the only way to talk to Ceph-RGW. `skip_s3_checksum` is the single flag there with a real integrity cost, and it'd be a sharper finding on its own.

---

## Verdict

Ship it once the `kubernetes` constraint is bounded (`main.tf:7-10`). The other two are optional polish, and the ADR-0055 tension is worth a conversation but not a change here.

---
*Re-review by Claude Opus 5 at @stephane-segning's request. Fixes confirmed against the diff at `a7c0193`; the validate job log, the #937 `share_level` reference, Terraform v1.15.8, and #913's merge were each checked directly rather than taken from the summary.*

.coder/templates/poc-ubuntu/main.tf
```diff
@@ -0,0 +1,83 @@
...
+  required_providers {
+    coder = {
+      source  = "coder/coder"
+      version = "~> 2.13"
+    }
+    kubernetes = {
+      source  = "hashicorp/kubernetes"
+      version = ">= 2.23"
+    }
+  }
```
stephane-segning, line 11: **One constraint got left unbounded — and it's the one that actually crossed a major version.**

`coder` is now `~> 2.13` ✅ and `coderd` is `~> 0.0.3` ✅ (correct form for a `0.0.x` provider — that reads as `>= 0.0.3, < 0.1.0`). But `kubernetes = ">= 2.23"` is still open-ended, and look what it resolved to in the committed lock:

```hcl
provider "registry.terraform.io/hashicorp/kubernetes" {
  version     = "3.2.1"
  constraints = ">= 2.23.0"
}
```

A **v2 → v3 major jump**, taken by resolution rather than by anyone choosing it. The lock file makes today reproducible, so nothing is broken right now — but the constraint still admits v4 the moment anyone runs `terraform init -upgrade` or regenerates the lock, and the whole point of this PR is that the delivery is reproducible from git.

```hcl
kubernetes = {
  source  = "hashicorp/kubernetes"
  version = "~> 3.2"
}
```

Worth a moment's thought before pinning: v3 was a major, so if `kubernetes_pod_v1` semantics changed you want to know now rather than at the first workspace build. The CI `validate` only proves it parses — the provisioner is where it runs.

terraform/coder/main.tf
```diff
@@ -0,0 +1,37 @@
...
+
+    # S3-compatible (Ceph-RGW) backend flags.
+    endpoints                   = { s3 = "https://nbg1.your-objectstorage.com" }
+    use_path_style              = true
+    use_lockfile                = true # verify Ceph-RGW conditional-write support
+    skip_s3_checksum            = true # verify; AWS SDK v2 checksums can break S3-compatible uploads
+    skip_credentials_validation = true
+    skip_requesting_account_id  = true
+    skip_metadata_api_check     = true
+  }
```
stephane-segning, line 31: Both `verify` caveats survived the rewrite, which is the right call — but worth noticing that **the first deploy on `main` is the verification**, and there's no documented fallback if it fails.

- `use_lockfile = true` needs Ceph-RGW to honour conditional writes (`If-None-Match`). If it doesn't, this fails at `init`/`plan` — so it fails *safe*, which is good. But someone will hit it at 2am with no note saying what to do.
- `skip_s3_checksum = true` is the one flag here with a genuine tradeoff (it disables the AWS SDK v2 integrity check). Worth keeping — it's usually required against S3-compatible stores — but it's the one to name if state ever looks corrupt.

Suggest one line in `README.md`: if `use_lockfile` isn't honoured, the fallback is `use_lockfile = false` **plus** an explicit note that deploys are then unserialized at the state layer (the workflow's `concurrency` group still serializes them, which is most of the protection anyway).

On the other review's shared-bucket point: I agree that's worth carrying forward. This state lives in `ssegning-k8s-state` alongside CNPG backups and the LGTM blocks, so one leaked key has platform-wide reach. Not a blocker for a pilot, but a separate bucket with scoped credentials is the right end state — worth an issue rather than scope creep here.

(`skip_credentials_validation` / `skip_metadata_api_check` are simply required for a non-AWS endpoint — no tradeoff to weigh there.)

.github/workflows/validate-coder-templates.yml
```diff
@@ -0,0 +1,53 @@
...
+          terraform validate
+
+      - name: Validate template sources
+        run: |
+          terraform fmt -check -recursive .coder/templates
+          for d in .coder/templates/*/; do
+            echo "validating $d"
+            terraform -chdir="$d" init -backend=false
+            terraform -chdir="$d" validate
+          done
```
stephane-segning: ✅ This is the fix I asked for, and I verified it actually runs rather than trusting the green check — the job log shows it installing the locked providers and validating for real:

```
validating .coder/templates/poc-ubuntu/
- Installing coder/coder v2.18.0...
- Installing hashicorp/kubernetes v3.2.1...
Success! The configuration is valid.
```

Good: the loop is glob-driven, so the next template is covered automatically with no workflow edit.

The remaining gap — flagged in the other review too — is that there's still no `terraform plan`, so `versions[].directory` resolving wrong, or a backend/state problem, surfaces only after merge. I'd push back on the "dry backend" phrasing though: a plan that doesn't read the real state can't tell you whether a new version will be created, which is the thing you actually want to see. If this is worth closing, the real shape is a **read-only plan against real state** — a second S3 key or a scoped credential — and that's a design decision, not a workflow tweak.

Given the deploy job plans immediately before applying and `coderd_template` changes are additive (a new version, not a destroy), I'd call the current gate proportionate for a pilot. Worth revisiting when this manages templates people depend on.

terraform/coder/main.tf
```diff
@@ -0,0 +1,37 @@
...
+  # Modern (Terraform >= 1.6) flag names: `endpoint` -> `endpoints.s3`,
+  # `force_path_style` -> `use_path_style`. `use_lockfile` gives S3-native state
+  # locking; verify Ceph-RGW honours conditional writes (If-None-Match).
+  backend "s3" {
+    bucket = "ssegning-k8s-state"
+    key    = "coder/templates/terraform.tfstate"
+    region = "us-east-1"
+
+    # S3-compatible (Ceph-RGW) backend flags.
+    endpoints                   = { s3 = "https://nbg1.your-objectstorage.com" }
```
Dericko681, line 24: does this object store already exists on the ssegning-aws?

terraform/coder/main.tf
```diff
@@ -0,0 +1,37 @@
...
+    skip_s3_checksum            = true # verify; AWS SDK v2 checksums can break S3-compatible uploads
+    skip_credentials_validation = true
+    skip_requesting_account_id  = true
+    skip_metadata_api_check     = true
+  }
+}
+
+provider "coderd" {
+  url   = var.coder_url
+  token = var.coder_token
```
Dericko681, line 36: so this CI has a coder account and hence coder_token to auth as itself?

.github/workflows/deploy-coder-templates.yml
```diff
@@ -0,0 +1,69 @@
+name: Deploy Coder Templates
+
+# GitOps apply for the declarative Coder template delivery (ADR-0124 — see
+# terraform/coder/README.md): on merge to main, reconcile the `coder` `coderd`
```
Dericko681, line 4: if the validate-coder-templates workflow fails, will this deploy run?

Dericko681 (changes requested):
I'm a little behind this PR pls speed me up

terraform/coder/main.tf
benie-joy-possi: yes it does

terraform/coder/main.tf
benie-joy-possi: the coder url here is coder.ai.camer.digital and the token will be generated for this url

benie-joy-possi: so this hashicorp coderd plugin is built to be able to connnect to a coder deployed and apply the templates defined

PR #1000, opened 2026-08-13
chore(observability): Improve grafana for speed and UX

### LINKED ISSUE #988: [Ticket]: Investigate how to speed up the grafana
### Type

Spike / Investigation

### Summary

We need to investigate and document the causes of the poor performance of grafana, document the possible fixes to the problem and implement the best fix

### Intent

To fix the problem of grafana dashboards taking too long to load

### Source of truth (links)

https://github.com/ADORSYS-GIS/ai-helm

### Current Behavior

- dashboards take too long to load

### Expected Behavior

- dashboards take significantly lesser time to load improving user experience

### Acceptance Criteria

- dashboard loads faster

### Out of Scope

_No response_

### Technical Context

_No response_

### Risks

_No response_

### Test Plan

_No response_

### Verification evidence

To be provided

### Human accountable owner

@koufan

### AI Usage Declaration

Not used

### Human verification completed

- [ ] I understood the intent
- [ ] I checked the source of truth
- [ ] I reviewed all AI-generated text/code
- [ ] I verified the implementation manually
- [ ] I verified the tests
- [ ] I checked for hallucinated assumptions
- [ ] I documented remaining risks
- [x] I am the accountable owner and accept responsibility for this ticket.

Dericko681:
## 1. Summary

This PR changes:

- rewrites the dashboards to read data from mimir instead of loki to increase speed. This problem was documented in [ADR 0058](https://github.com/ADORSYS-GIS/ai-helm/blob/main/docs/adr/0058-precompute-gateway-usage-metrics-to-mimir.md) and implementation completed here

Prerequisite changes have been merged [here](https://github.com/ADORSYS-GIS/ai-helm-values/commit/893ef04d3c0df3d12b596e13092861b8d62c80d8)

It solves:

- Dashboards loading very slowly leading to bad user experience
- Default time span not set to `This month only` leading to misguided info for non-technical users
- Grafana alerts using "last 30 days" instead of "this month so far" leading to false "over budget" alarms

documented in #988 

---

## 2. Intent

The intent of this PR is:

> To solve the problems listed above

---

### In Scope

- [Included change 1]
- [Included change 2]

### Out of Scope

- [Excluded change 1]
- [Excluded change 2]

---

## 4. Verification

I verified this change by:

- [ ] Running automated tests
- [x] Running manual tests
- [ ] Checking logs
- [ ] Checking metrics
- [ ] Testing error cases

Commands run:

```bash
open dashboards on grafana
```

Results:

```text
[[paste result or link]](https://grafana.ai.camer.digital/d/envoy-ai-gateway-chats-by-user-test/ai-gateway-e28094-chats-by-user-test?orgId=1&from=now%2FM&to=now&timezone=browser&var-user=selastlambou@gmail.com&refresh=30s)
```

---

## 5. Screenshots / Evidence
image

* Screenshot: [link]
* Logs: [link]
* Metrics: [link]
* Recording: [link]

---

## 6. Risk Assessment

Risk level:

* [ ] Low
* [x] Medium
* [ ] High

Potential risks:

* [Risk 1]
* [Risk 2]

Mitigation:

* [Mitigation 1]
* [Mitigation 2]

---

## 7. AI Usage Declaration

AI was used for:

* [ ] Generating code
* [ ] Refactoring
* [ ] Generating tests
* [ ] Drafting documentation
* [ ] Reviewing the diff
* [x] Not used

Human verification:

* [ ] I understand every meaningful change in this PR
* [ ] I checked generated code manually
* [ ] I removed unsupported AI assumptions
* [ ] I accept responsibility for this PR

---

## 8. Reviewer Focus

Please focus your review on:

* [x] Correctness
* [ ] Architecture
* [ ] Security
* [ ] Performance
* [ ] Tests
* [ ] Maintainability
* [ ] Product intent
* [ ] Edge cases


charts/observability-dashboards/values.yaml
```diff
@@ -589,13 +589,13 @@ alerting:
         - uid: cost-monthly-burn
           title: AI Gateway — monthly burn high
           datasourceUid: mimir
-          expr: sum(increase(loki_process_custom_gen_ai_usage_cost_micro_usd[30d]))/1e6
+          expr: sum(ai_gateway_cost_mtd_total_usd)/1e6
```
leghadjeu-christian: **P0 — Mimir rejects this config outright, and the expressions do not parse.**

> **Corrected.** My original first point claimed the prerequisite was unmerged. @Dericko681 is right — ai-helm-values#244 has since merged, so `ruler.enabled: true` and both recording rules are on `main`. That point is withdrawn. The two below are unaffected, and the merge makes the first one **more** urgent, not less: the config is now on `main`, pending sync.

**1. Mimir has no `ruler.rules` field — this crashloops the whole stack, not just the ruler.**

The rules sit at `mimir.structuredConfig.ruler.rules`. From the chart source (`mimir-distributed-5.8.0.tgz`):

```gotemplate
{{/* templates/_helpers.tpl:136-138 */}}
{{- define "mimir.calculatedConfig" -}}
{{ tpl (mergeOverwrite (include "mimir.unstructuredConfig" . | fromYaml) .Values.mimir.structuredConfig | toYaml) . }}
```

`structuredConfig` is merged into `mimir.yaml`, rendered into a single ConfigMap (`templates/mimir-config.yaml`, `configStorageType: ConfigMap` by default), and mounted by **20 component templates** — distributor, ingester, querier, store-gateway, compactor, query-frontend among them — each passing `-config.file=/etc/mimir/mimir.yaml`.

Tested against the exact deployed version (chart `5.8.0` → appVersion `2.17.0`, resolved from the Grafana chart index):

```console
$ docker run --rm -v "$PWD":/w:ro grafana/mimir:2.17.0 -config.file=/w/mimir-config.yaml -target=ruler
error loading config from /w/mimir-config.yaml: Error parsing config file: yaml: unmarshal errors:
  line 3: field rules not found in type ruler.Config
```

The identical file with only the `rules:` key removed starts cleanly. This matches upstream design — Mimir has no config-file equivalent of Prometheus's `rule_files:`; rule groups load from `-ruler-storage.backend` via `mimirtool rules`, the ruler HTTP configuration API, or `mimir-rules-action`. Ref: [Mimir ruler docs](https://grafana.com/docs/mimir/latest/references/architecture/components/ruler/).

I checked the alternative delivery paths too: `extraObjects` is absent from `mimir.yaml`, that file is the only one in the values repo containing `record:`, and neither repo's CI runs `mimirtool` or `mimir-rules-action`.

**2. The expressions do not parse.** Both Prometheus `promtool` 3.13.2 and Grafana's own `mimirtool` reject them:

```text
promtool check rules  → 5:7: parse error: unexpected "(" in @, expected timestamp
mimirtool rules check → 5:7: parse error: unexpected "(" in @, expected timestamp
```

`@` takes a float literal (or `start()`/`end()`), never an expression: *"The time supplied to the `@` modifier is a Unix timestamp and described with a float literal"* — [Prometheus querying basics](https://prometheus.io/docs/prometheus/latest/querying/basics/#modifier).

---

**Why this lands silently rather than loudly.** `templates/alerting.yaml:136` defaults `noDataState: OK`, and none of the five rules override it. A query on a metric with no series returns empty → No Data → Normal → permanently green ([Grafana alert state docs](https://grafana.com/docs/grafana/latest/alerting/fundamentals/alert-rules/state-and-health/)). That is what this file warns about at line 345:

> *"an alert on a metric that does not exist is worse than no alert, because noDataState: OK makes it permanently, invisibly green"*

The house fix used at lines 356 / 498 / 540 is `or vector(0)`.

**Suggestion:** drop the alerting change from this PR. The dashboard half (`now/M` + `increase([$__range])`) already delivers calendar-MTD with no ruler at all.

charts/observability-dashboards/values.yaml
```diff
@@ -612,48 +612,50 @@ alerting:
           datasourceUid: mimir
           expr: >-
             sum by (email, display_name)
-              (increase(loki_process_custom_gen_ai_usage_cost_micro_usd{billing_plan="free"}[30d])) / 1e6
+              (ai_gateway_cost_mtd_usd{billing_plan="free"}) / 1e6
```
leghadjeu-christian: **P1 — even once the rules load, counter subtraction is wrong for this counter.**

The proposed shape is `sum(X) - sum(X @ month_start)`. That ignores counter resets, and this counter resets constantly: it carries `max_idle_duration = "2h"` (ai-helm-values `environments/prod/values/alloy.yaml:518-523`), documented as *"Maximum amount of time to wait until the metric is marked as 'stale' and removed"* ([Alloy `loki.process` docs](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/)). Each user's series is deleted after 2h idle and restarts at zero — several times a day for anyone who isn't continuously active.

promtool unit test, same expression shape with the `@` argument replaced by a literal so it parses:

```yaml
# alice: 0→500, series pruned, restarts 0→120   TRUE month-to-date = 620
# bob:   first sample lands AFTER month start
values: '0 100 200 300 400 500 0 60 120'
values: '_ _ _ 0 40 80 120 160 200'
```
```text
sum by (email) (cost) - sum by (email) (cost @ 0)  → {email="alice"} 120
sum by (email) (increase(cost[9m]))                → {email="alice"} 620, {email="bob"} 200
sum(cost) - sum(cost @ 300)                        → {} -260
```

Three confirmed defects:

- alice under-reports by **81%** (120 vs 620);
- **bob is dropped entirely** — no series on the right of the subtraction means no result at all, so any user whose first request lands after month start never alerts. That is the population you most want this rule to catch;
- the value can go **negative**.

The expression being replaced used `increase()`, which per the [Prometheus function docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase) *"Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for."* Whatever replaces this needs to keep that property.

charts/observability-dashboards/values.yaml
```diff
@@ -612,48 +612,50 @@ alerting:
...
+            which resets on a schedule, not calendar-month) — someone can be over
             budget here while their LIVE enforcement budget is well under cap.
             Check the AI Gateway — rate-limit quota dashboard for the live state.
         - uid: cost-actor-budget-free
           title: AI Gateway — free-tier actor well over budget ($30)
           datasourceUid: mimir
           expr: >-
             sum by (email, display_name)
-              (increase(loki_process_custom_gen_ai_usage_cost_micro_usd{billing_plan="free"}[30d])) / 1e6
+              (ai_gateway_cost_mtd_usd{billing_plan="free"}) / 1e6
```
leghadjeu-christian: **P1 — the `_usd` metrics carry micro-USD.**

The recording rules are `sum(loki_process_custom_gen_ai_usage_cost_micro_usd)` with no division — verified, zero occurrences of `1e6` in the rule file. Every consumer divides afterwards instead (here and at 592, 615, 667, 701).

The arithmetic is self-consistent, so no wrong numbers today. But `ai_gateway_cost_mtd_usd` naming a micro-USD value is exactly the 1,000,000× trap that `_shared.usd()` exists to prevent, and `per_user.py`'s own docstring spells it out:

> *"Every cost panel must divide by 1e6 before applying the currencyUSD unit, or values display 1,000,000x too large."*

The next person to write a panel against this metric will read the name, not the rule. Either rename to `ai_gateway_cost_mtd_micro_usd`, or fold the `/1e6` into the recording rule and drop it from all four alert expressions.

charts/observability-dashboards/values.yaml
```diff
@@ -695,7 +698,7 @@ alerting:
           datasourceUid: mimir
           expr: >-
             (sum(provider_billing_total_cost_micro_usd) / 1e6)
-            / clamp_min(sum(increase(loki_process_custom_gen_ai_usage_cost_micro_usd[30d])) / 1e6, 1)
+            / clamp_min(sum(ai_gateway_cost_mtd_total_usd) / 1e6, 1)
```
leghadjeu-christian: **P1 — the reconciliation alert is the fifth casualty, and it fails closed-green too.**

`clamp_min(sum(ai_gateway_cost_mtd_total_usd) / 1e6, 1)` — when the metric has no series, `sum()` returns empty, `clamp_min` of empty is empty, and the whole division is empty. With `noDataState: OK` (default, `templates/alerting.yaml:136`) the provider-vs-gateway drift alert goes permanently green as well. Worth calling out separately because this one is the invoice safety net, not a budget nudge.

**P2, while you're here —** all four cost rules keep `fromSeconds: 2592000` while their descriptions now say month-to-date. It's harmless in practice, since `templates/alerting.yaml:160` sets `instant: true, range: false` and the relative range is ignored for the query — but it's a stale 30-day claim left in the config for the next reader.

tools/dashboards/src/dashboards/envoy_ai_gateway/my_usage.py
```diff
@@ -232,46 +126,27 @@ def _panel_total_tokens() -> stat.Panel:
...
     # Percent of the editable monthly budget spent in the selected range.
-    # Same Loki cost query as _panel_total_cost, divided by the $budget
+    # Same Mimir cost query as _panel_total_cost, divided by the $budget
     # textbox variable (default $15 — the free-plan monthly budget).
-    expr = (
-        "100 * ("
-        f"sum(sum_over_time({_SELECTOR} {_unwrap('gen_ai_usage_custom_total_cost')} [$__range]))"
-        " / 1e6) / $budget"
-    )
+    expr = f"100 * (sum(increase({METRIC_COST_MICRO_USD}{_SEL}[$__range])) / 1e6) / $budget"
```
leghadjeu-christian, line 133: **P2 — two notes on this gauge now that the dashboard time is `now/M → now`.** Neither blocks; both are worth a line in the panel description.

**1. Blank for the first minutes of each month.** Just after midnight on the 1st, `$__range` is tens of seconds, and `increase()` over a window shorter than the sample interval returns nothing. Verified on a 60s interval:

```text
sum(increase(cost[30s]))  eval 3m  → nil   (empty, no data)
sum(increase(cost[5m]))   eval 3m  → 300
```

This affects every headline stat on the Mimir-backed boards, not just this gauge. `docs/patterns/cost-observability.md` already carries the ≥2-samples warning for the daily bars — the same trap now applies at month rollover. A `[$__range:]` subquery or a `max($__range, 5m)` floor would close it.

**2. This panel moves from exact to estimated.** The old Loki `sum_over_time` was a true sum; `increase()` *"is extrapolated to cover the full time range as specified in the range vector selector, so that it is possible to get a non-integer result even if a counter increases only by integer increments"* ([Prometheus docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#increase)). For sparse users the gauge can disagree with both the invoice and the alert threshold.

That trade-off is inherent to ADR-0058 and not introduced by this PR — but this is the panel where a user compares their own number against `$budget` (default $15), so it is the one place the discrepancy will get noticed and questioned.

leghadjeu-christian (changes requested):
> **Edited to correct one point.** My original summary said the prerequisite recording rules were unmerged. @Dericko681 is right that ai-helm-values#244 has since merged — that claim is withdrawn. The remaining blockers are unaffected, and the merge makes them more urgent rather than less, since the config is now on `main`.

Reviewed against the live `ai-helm-values` state, with every claim validated by executing the tool rather than reading docs — `promtool` 3.13.2, Grafana `mimirtool`, and `grafana/mimir:2.17.0` (the exact version mimir-distributed 5.8.0 deploys).

## The dashboard migration is good and should ship

`_shared.py` is a clean consolidation, and the sharp edges are documented where they bite: the instant-vs-range `$__range` substitution difference in `prom_target`, the `rate()[5m]*60` choice over `increase()[1m]` with the ≥2-samples reason stated inline. Panels that genuinely need log data — `response_code`, the per-request listing, missing-cost — were correctly left on Loki and the docs table now says so per-dashboard. I checked two things that looked like regressions in the diff and are not: template variables keep `refresh=2` (on time range change) with the datasource correctly flipped `loki` → `prometheus`, and the drift check passes so the committed JSON matches the generators. The latency panel is honestly titled "p95 latency (5m)" rather than implying it follows the range picker, and the 100ms–120s buckets fit LLM latency well.

## The alerting change should not merge in this state

Two blockers, detailed inline on line 592:

1. the rules sit at `mimir.structuredConfig.ruler.rules`, a field Mimir does not have — `grafana/mimir:2.17.0` **refuses to start** with `field rules not found in type ruler.Config`. Because the chart merges `structuredConfig` into a single `mimir.yaml` ConfigMap mounted by 20 component templates, that crashloops the whole Mimir stack, not just the ruler. Mimir loads rules from ruler storage, not the config file;
2. the expressions **fail to parse** in both `promtool` and `mimirtool` — `@` takes a float literal, not an expression.

Because `noDataState` defaults to `OK`, none of this surfaces as a broken alert. It surfaces as a permanently green one. This file already states the standard at line 345 — *"an alert on a metric that does not exist is worse than no alert"* — with `or vector(0)` as the house fix at 356/498/540. As written, #988's false "over budget" alarm would be replaced by never alarming at all, across all four budget rules plus the provider-billing reconciliation net.

Separately, even once the rules load, `sum(X) - sum(X @ month_start)` is the wrong shape for a counter that Alloy prunes every 2h — reproduced inline on line 615: an 81% under-report, new users dropped entirely, and negative values.

## Suggested path

Split the PR. Ship the dashboards now — `now/M` + `increase([$__range])` delivers calendar-MTD with no ruler at all, and that is the user-visible half of #988. Land the alerting change separately once the rules parse, survive counter resets, and live in ruler storage. Worth noting that Grafana alert rules have no calendar-aligned range, so a rolling window may honestly be the right answer for enforcement — that is a design question worth settling before the next attempt.

Minor: the PR template's Scope, Risk and Mitigation sections are still placeholders on a change that rewires all budget alerting.

Everything above is reproducible; happy to share the exact rule/test files if useful.

charts/observability-dashboards/values.yaml
Dericko681: It exists

Dericko681: the PR was still open https://github.com/ADORSYS-GIS/ai-helm-values/pull/244

leghadjeu-christian: You're right, and thanks for the pointer — ai-helm-values#244 has merged since I wrote that, so `ruler.enabled: true` and both recording rules are on `main` now. I've withdrawn that point from the comment above.

The two remaining blockers are unaffected, and the merge makes them more urgent rather than less, because the config is now on `main` pending sync:

- `grafana/mimir:2.17.0` (what chart 5.8.0 deploys) refuses to start on this config: `field rules not found in type ruler.Config`. Mimir has no `ruler.rules` config field — rule groups load from ruler storage, not `mimir.yaml`. Since the chart merges `structuredConfig` into one ConfigMap mounted by 20 components, that is a stack-wide crashloop, not just the ruler.
- The expressions don't parse in either `promtool` or `mimirtool`: `parse error: unexpected "(" in @, expected timestamp`. PromQL's `@` takes a float literal, not an expression.

Full evidence in the edited comment above. Happy to be wrong again if you can get the ruler to load these — that would be the quickest way to settle it.

Issue #161, opened 2026-04-08
[Ticket]: Add SBOM generation and cosign signing to the release pipeline


stephane-segning:
&gt; Upgraded to the AI-Helm governance **Dev Ticket** format · https://adorsys-gis.github.io/ai-governance/03-dev-ticket-template

## 1. Type
- [x] Feature / Pipeline change  _(extends the release path with SBOM + signing)_

## 2. Summary
Extend the supply-chain controls **beyond basic scanning**: produce an **SBOM** for release artifacts
and **sign** them with **cosign**, so release outputs are easier to trust and verify.

## 3. Intent
&gt; Scan gates catch known issues before release; SBOM + cosign make what ships **inventoriable and
&gt; verifiable** afterward. This ticket defines or implements that next stage of release hardening.

## 4. Source of Truth
- Parent user story: #136 · Parent epic: #144
- **CI (current, ADR-0055 continuous delivery):** [`.github/workflows/publish-charts-oci.yml`](.github/workflows/publish-charts-oci.yml) — the live release path (charts → OCI on merge to `main`). This is where SBOM + cosign now slot in.
- **Design / why:** [ADR-0130](docs/adr/0130-sbom-and-cosign-sign-oci-charts.md) (Proposed) — keyless cosign + rendered-image SBOM (child of this ticket).
- **Verification / how:** [`docs/patterns/supply-chain.md`](docs/patterns/supply-chain.md).
- _Historical note:_ this ticket was originally drafted against the now-retired tag-based release (ADR-0031, `release-helm-charts.yml`, `docs/releasing.md`). That model was superseded by ADR-0055 (continuous delivery, 2026-06-22); `docs/releasing.md`/`tools/release.sh` are retired. SBOM + signing live in the **OCI publish** workflow.
- Scan gates (already landed, sibling #160): `.github/workflows/security.yml` → `ADORSYS-GIS/ai-governance/security-gates.yml` (Trivy + Gitleaks); Trivy config scan of rendered manifests in `release-helm-charts.yml`.

## 5. Expected Behavior / Deliverables
- SBOM generation wired into the release path: per-chart CycloneDX inventory of the container images the chart's rendered manifests deploy (`tools/oci-chart-sbom.sh`), attached via `cosign attach sbom`.
- Keyless cosign signing of each published chart OCI artifact (Sigstore OIDC).
- Documented verification expectations for signed artifacts (`docs/patterns/supply-chain.md`).

[...]

PR #1028, opened 2026-08-19
fix(ci): drop oci:// scheme for cosign attach/sign (#161)

Guy-Ghis:
## 1. Summary

This PR changes:

- `.github/workflows/publish-charts-oci.yml` — add a scheme-less `COSIGN_REPO` (`ghcr.io/adorsys-gis/charts`) env var and use it for the `cosign attach sbom` and `cosign sign` commands (the helm-only `OCI_REPO` keeps its `oci://` prefix).

It solves:

- Bug in **#161** (SBOM + cosign signing for OCI charts): the `Sign charts (cosign) + attach SBOM` step failed in the first manual workflow run with `Error: Get "https://oci/v2/": dial tcp: lookup oci ... server misbehaving`. cosign does not support the `oci://` scheme, so it parsed `oci` as the registry hostname.

---

## 2. Intent

The intent of this PR is:

> cosign does not understand the `oci://` scheme that helm requires, so passing `$OCI_REPO` (which is `oci://ghcr.io/adorsys-gis/charts`) to `cosign attach sbom` / `cosign sign` made cosign treat `oci` as the registry host and fail with a DNS lookup error (`Get "https://oci/v2/"`). This change introduces a scheme-less `COSIGN_REPO` for the cosign commands, leaving `OCI_REPO` untouched for helm. Helm requires the scheme; cosign must not have it.

---

### In Scope

- Add `COSIGN_REPO: ghcr.io/adorsys-gis/charts` to the workflow `env`.
- Switch `cosign attach sbom` and `cosign sign` to reference `$COSIGN_REPO/$c:$ver` instead of `$OCI_REPO/$c:$ver`.

### Out of Scope

- Migrating from the deprecated `cosign attach sbom` to SBOM attestations (`cosign attest --predicate`) — a separate, larger change (different verify syntax); tracked as a follow-up, not part of this fix.
- The benign `cosign login` "credentials stored unencrypted" warning (ephemeral runner, short-lived scoped `GITHUB_TOKEN`) — no change needed.

---

## 4. Verification

I verified this change by:

- [x] Running automated tests
- [x] Running manual tests
- [ ] Checking logs
- [ ] Checking metrics
- [ ] Testing error cases

Commands run:

```bash
# workflow YAML is well-formed (mikefarah yq, as used in CI)
yq e '.' .github/workflows/publish-charts-oci.yml >/dev/null

# confirm only cosign commands use the scheme-less repo; helm keeps oci://
grep -n "cosign" .github/workflows/publish-charts-oci.yml
grep -n "OCI_REPO\|COSIGN_REPO" .github/workflows/publish-charts-oci.yml
```

Results:

```text
YAML OK
cosign login "$REGISTRY"          # REGISTRY is already scheme-less (ghcr.io)
cosign attach sbom ... "$COSIGN_REPO/$c:$ver"   # fixed
cosign sign --yes "$COSIGN_REPO/$c:$ver"        # fixed
OCI_REPO: oci://ghcr.io/adorsys-gis/charts       # helm unchanged
COSIGN_REPO: ghcr.io/adorsys-gis/charts          # new
```

> Root cause reproduced from the failing run logs: `Get "https://oci/v2/"` — cosign resolving `oci` as the registry hostname because the reference carried the `oci://` scheme. The fix removes the scheme only for cosign. Final proof is a successful `publish-charts-oci` run; re-run the workflow after merge.

---

## 5. Screenshots / Evidence

* The failing step's error (from the first manual run of `publish-charts-oci`):

```text
Error: Get "https://oci/v2/": dial tcp: lookup oci on 127.0.0.53:53: server misbehaving
error during command execution: Get "https://oci/v2/": ...
Error: Process completed with exit code 1.
```

* The fix: `COSIGN_REPO` env var added; both cosign commands switched to `$COSIGN_REPO/$c:$ver`.
* Pending (post-merge): a successful `publish-charts-oci` run with `signed <chart> → <ver>` in the logs.

---

## 6. Risk Assessment

Risk level:

* [x] Low
* [ ] Medium
* [ ] High

Potential risks:

* `COSIGN_REPO` and `OCI_REPO` could drift apart (they must differ only by the scheme).

Mitigation:

* Comment on `COSIGN_REPO` states it "must match OCI_REPO minus the scheme", so a drift is visible at review time.

---

## 7. AI Usage Declaration

AI was used for:

- [x] Understanding existing code
- [ ] Generating code
- [x] Refactoring
- [ ] Generating tests
- [ ] Drafting documentation
- [x] Reviewing the diff
- [ ] Not used

Human verification:

- [x] I checked generated tests manually
- [x] I removed unsupported AI assumptions

---

## 8. Reviewer Focus

Please focus your review on:

- [x] Correctness
- [x] Architecture
- [ ] Security
- [ ] Performance
- [ ] Tests
- [ ] Maintainability
- [x] Product intent
- [x] Edge cases

### DIFF
--- a/.github/workflows/publish-charts-oci.yml
+++ b/.github/workflows/publish-charts-oci.yml
@@ -40,6 +40,10 @@ concurrency:
 env:
   REGISTRY: ghcr.io
   OCI_REPO: oci://ghcr.io/adorsys-gis/charts
+  # Scheme-less repo for cosign: cosign does NOT support the `oci://` scheme
+  # (it would treat `oci` as a registry hostname → "Get https://oci/v2/" DNS
+  # failure), unlike helm which requires it. Must match OCI_REPO minus the scheme.
+  COSIGN_REPO: ghcr.io/adorsys-gis/charts
   # Local library / template charts: built first (dependency-ordered) so their
   # file:// deps are vendored into dependents, and NEVER published standalone
   # (consumers vendor them at package time). `apps` is the root umbrella — the
@@ -212,7 +216,7 @@ jobs:
             if tools/oci-chart-sbom.sh "charts/$c" "sbom-$c" "$sbom" >/dev/null 2>&1 \
                && [[ -s "$sbom" ]] && jq -e . "$sbom" >/dev/null 2>&1; then
               echo "  attaching SBOM: $(jq -c '{components:(.components|length)}' "$sbom")"
-              cosign attach sbom --sbom "$sbom" "$OCI_REPO/$c:$ver" || \
+              cosign attach sbom --sbom "$sbom" "$COSIGN_REPO/$c:$ver" || \
                 echo "  (warn) sbom attach failed for $c"
             else
               echo "  (skip) no renderable SBOM for $c — signing artifact only"
@@ -225,7 +229,7 @@ jobs:
             # only marks the run red so the operator knows the chart was not
             # signed and can re-run (idempotent). If this ever starts blocking
             # merges in practice, revisit and downgrade to a warn like the SBOM.
-            cosign sign --yes "$OCI_REPO/$c:$ver"
+            cosign sign --yes "$COSIGN_REPO/$c:$ver"
             echo "  signed $c → $ver"
           done
 
--- a/.github/workflows/publish-charts-oci.yml
+++ b/.github/workflows/publish-charts-oci.yml
@@ -40,10 +40,6 @@ concurrency:
 env:
   REGISTRY: ghcr.io
   OCI_REPO: oci://ghcr.io/adorsys-gis/charts
-  # Scheme-less repo for cosign: cosign does NOT support the `oci://` scheme
-  # (it would treat `oci` as a registry hostname → "Get https://oci/v2/" DNS
-  # failure), unlike helm which requires it. Must match OCI_REPO minus the scheme.
-  COSIGN_REPO: ghcr.io/adorsys-gis/charts
   # Local library / template charts: built first (dependency-ordered) so their
   # file:// deps are vendored into dependents, and NEVER published standalone
   # (consumers vendor them at package time). `apps` is the root umbrella — the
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-charts-oci.yml
```diff
@@ -40,6 +40,10 @@ concurrency:
 env:
   REGISTRY: ghcr.io
   OCI_REPO: oci://ghcr.io/adorsys-gis/charts
+  # Scheme-less repo for cosign: cosign does NOT support the `oci://` scheme
+  # (it would treat `oci` as a registry hostname → "Get https://oci/v2/" DNS
+  # failure), unlike helm which requires it. Must match OCI_REPO minus the scheme.
+  COSIGN_REPO: ghcr.io/adorsys-gis/charts
```
leghadjeu-christian: Nit (non-blocking): `COSIGN_REPO` is a hand-duplicated literal of `OCI_REPO` minus the scheme, rather than derived from it. If `OCI_REPO` ever changes, only updating that line would let `COSIGN_REPO` silently point at the old location — the comment is the only guard against drift. Deriving it instead (e.g. `COSIGN_REPO="${OCI_REPO#oci://}"` computed in a step) would make drift structurally impossible.

Low severity: a real drift would likely surface loudly anyway, since `cosign sign` is a deliberate hard-fail per the comment below.

leghadjeu-christian (approved):
Correct, narrowly-scoped fix. Confirmed the root cause against current `main`: cosign doesn't parse the `oci://` scheme and was resolving `oci` as a hostname, while `helm` genuinely needs the scheme — so keeping `OCI_REPO` for helm and adding a scheme-less `COSIGN_REPO` for the two `cosign` calls is the right split. Verified no other workflow has the same pattern needing the same fix.

One minor, non-blocking suggestion left inline.

.github/workflows/publish-charts-oci.yml
Guy-Ghis: Good catch — addressed. `COSIGN_REPO` is no longer a hand-duplicated literal. It's now removed from the workflow-level `env` and derived in the sign step:

```yaml
COSIGN_REPO="${OCI_REPO#oci://}"
```

This strips the `oci://` prefix from `OCI_REPO` in-shell, so the two can't drift structurally (e.g. if `OCI_REPO` changes, `COSIGN_REPO` follows automatically). Pushed to `fix/161-cosign-oci-scheme` (`f780df7d`); verified YAML is well-formed and `${OCI_REPO#oci://}` → `ghcr.io/adorsys-gis/charts`.

